PR #12, opened 2021-12-03
feature: add helper to check if all required labels are available

sumitanvekar:
### :pencil: Description
add helper to check if all required labels are available


action.yml
```diff
@@ -79,6 +79,9 @@ inputs:
   number_of_assignees:
     description: 'Number of people to assign a PR to'
     required: false
+    contains_all_labels:
+    description: 'A comma seperated list of labels used to check if all of them are present on a pull request'
+    required: false
```
danadajian, line 84: We should be able to use the existing `labels` input here.

danadajian (changes requested):

action.yml
sumitanvekar: i kept it separate since it gives more option in future. e.g if we want to add checks like "one of", "none of", etc.

danadajian: Actually, I just realized this can be accomplished within github workflows already

danadajian: https://docs.github.com/en/actions/learn-github-actions/expressions#contains

sumitanvekar:
I am closing this pr since it can be achieve with native gh actions support. Thanks Dan!

PR #25, opened 2021-12-07
feat: add create-project-card and move-project-card helpers

salvadorgonmo:
### :pencil: Description
This PR adds 2 helpers:

- create-project-card: This can create a GitHub Project card within a Project into the same repository the GitHub action is fired. It either can be created with the detailed information about the Pull Request, or it can be an specific text you want to provide via the `note` property in the `yml` file.

- move-project-card: This can move a project card (if exists) from a desired column to another one.
Example of a GitHub project:
Screen Shot 2021-12-07 at 18 17 15

You can also refer to this temporary PR to see it in action with a test project created in the same Repository scope: https://github.com/salvadorgonmo/github-helpers/pull/1


src/helpers/move-project-card.ts
```diff
@@ -0,0 +1,95 @@
...
+                });
+            }
+          });
+      }
+    })
+    .catch(error => {
+      if (error.status === 409) {
+        core.info('There was an error moving the project card.');
+      }
+    });
```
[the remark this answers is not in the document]
salvadorgonmo: Not really, removed it.

src/helpers/create-project-card.ts
```diff
@@ -0,0 +1,95 @@
...
+          .then(projects => {
+            const project = findProjectToModify(projects, project_name);
+            if (project) {
+              octokit.projects
+                .listColumns({
+                  project_id: project.id,
+                  per_page: 100
+                })
+                .then(response => {
+                  const filteredColumn = filterDestinationColumn(response, project_destination_column_name);
```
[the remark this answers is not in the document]
salvadorgonmo: Yeah, I created another file to reuse this similar code in both helpers. In order to successfully create a card, we need to place it under a column into the project, so it is a required field: https://docs.github.com/en/rest/reference/projects#create-a-project-card. However, I had a logic to move a card after the creation of it, but that could be done with a workflow too, so removed it.

src/helpers/create-project-card.ts
```diff
@@ -0,0 +1,95 @@
...
+                        })
+                        .then(response => {
+                          // move the card to the coulmn's bottom after created
+                          octokit.projects.moveCard({
+                            card_id: response.data.id,
+                            position: 'bottom',
+                            column_id: filteredColumn.id
+                          });
+                        });
+                    }
```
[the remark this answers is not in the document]
salvadorgonmo: There was no need actually to do that, so I removed it as I said in this comment: https://github.com/ExpediaGroup/github-helpers/pull/25#discussion_r765073202 Thank you for the suggestion

src/helpers/create-project-card.ts
```diff
@@ -0,0 +1,60 @@
...
+      column_id: filteredColumn?.id,
+      note,
+      ...context.repo
+    };
+  }
+
+  return {
+    column_id: filteredColumn?.id,
+    content_id: pullRequest.id,
+    content_type: 'PullRequest',
```
danadajian, line 56: Why do we want to include these params only if there is no "note"?

danadajian: Screen Shot 2021-12-08 at 11 34 02 AM
According to the [docs](https://octokit.github.io/rest.js/v18), it seems that all of these params are required for `createCard()`

src/helpers/move-project-card.ts
```diff
@@ -0,0 +1,55 @@
...
+
+  if (!columnsList) {
+    return null;
+  }
+
+  const destinationColumn = getDestinationColumn(columnsList, project_destination_column_name);
+  const originColumn = getOriginColumn(columnsList, project_origin_column_name);
+
+  if (!originColumn) {
+    return null;
```
danadajian, line 44: We could probably just log using `core.info()` that no origin column was found, and just `return`

src/helpers/create-project-card.ts
```diff
@@ -0,0 +1,60 @@
...
+    return {
+      column_id: filteredColumn?.id,
+      note,
+      ...context.repo
+    };
+  }
+
+  return {
+    column_id: filteredColumn?.id,
+    content_id: pullRequest.id,
```
danadajian, line 55: Just confirming, but for `content_id` do we need the `id` of the pull request, or could we just use the pull request number?

src/helpers/create-project-card.ts
```diff
@@ -0,0 +1,60 @@
...
+
+import { getDestinationColumn, getProjectColumns } from '../utils/get-project-columns';
+import { PullRequest } from '../types';
+import { context } from '@actions/github';
+import { octokit } from '../octokit';
+
+interface CreateProjectCardProps {
+  teams?: string;
+  pull_number: number;
+  login?: string;
```
danadajian, line 22: Seems like `login` is unused

src/utils/get-project-columns.ts
```diff
@@ -0,0 +1,40 @@
...
+  return octokit.projects.listColumns({ project_id: project.id, per_page: 100 });
+};
+
+const findProjectToModify = (projectsResponse: ProjectListResponse, project_name: string) =>
+  projectsResponse.data.find(project => project.name === project_name);
+
+export const getDestinationColumn = (columns: ColumnListResponse, project_destination_column_name: string) =>
+  columns.data.find(column => column.name === project_destination_column_name);
+
+export const getOriginColumn = (columns: ColumnListResponse, project_origin_column_name: string) =>
```
danadajian, line 39: I think this could be moved to `move-project-card.ts` since it is only used there

danadajian (commented):
Much more readable, thank you! Left some more comments

src/helpers/create-project-card.ts
salvadorgonmo: Not sure why the docs says that they are required, but if we send either `content_id` or `content_type` or both when we are also sending a note, it throws an error saying the following:
Screen Shot 2021-12-08 at 12 00 09

src/helpers/create-project-card.ts
salvadorgonmo: I did a try sending the pull number over the pull id, but looks like it is not working since it throws the following error:
Screen Shot 2021-12-08 at 12 22 43

Also, added some logs to see the differences between the `pull id` and the `pull number`. So the id is the winner 😅

src/helpers/create-project-card.ts
```diff
@@ -0,0 +1,64 @@
...
+  note?: string;
+}
+
+export const createProjectCard = async ({ pull_number, project_name, project_destination_column_name, note }: CreateProjectCardProps) => {
+  const getResponse = await octokit.pulls.get({ pull_number, ...context.repo });
+  const pullRequest = getResponse.data as PullRequest;
+  const columnsList = await getProjectColumns({ project_name });
+
+  if (!columnsList || columnsList.data.length === 0) {
+    core.info(`There are no columns associated to ${project_name} project.`);
```
danadajian, line 33: Do we want to fail the job here in this case? If so, we can use `core.error()`, but if not, we can leave as `core.info()`

src/helpers/create-project-card.ts
```diff
@@ -0,0 +1,64 @@
...
+
+  if (destinationColumn) {
+    return octokit.projects.createCard(cardParams);
+  } else {
+    core.info('No destination column was found');
+    return;
+  }
+};
+
+const generateCardParams = (note: string | undefined, filteredColumn: any, pullRequest: PullRequest) => {
```
danadajian, line 48: Could we have `note` be the last argument? Seems a bit weird to have the first arg be optional in the function signature.

```suggestion
const generateCardParams = (filteredColumn: any, pullRequest: PullRequest, note?: string) => {
```

src/helpers/create-project-card.ts
```diff
@@ -0,0 +1,64 @@
...
+  project_destination_column_name: string;
+  note?: string;
+}
+
+export const createProjectCard = async ({ pull_number, project_name, project_destination_column_name, note }: CreateProjectCardProps) => {
+  const getResponse = await octokit.pulls.get({ pull_number, ...context.repo });
+  const pullRequest = getResponse.data as PullRequest;
+  const columnsList = await getProjectColumns({ project_name });
+
+  if (!columnsList || columnsList.data.length === 0) {
```
danadajian, line 32:
```suggestion
  if (!columnsList?.data?.length) {
```

README.md
```diff
@@ -13,6 +13,13 @@ Each of the following helpers are defined in a file of the same name in `src/hel
 ### [**assign-pr-reviewers**](.github/workflows/assign-pr-reviewers.yml)
   * Randomly assigns members of a github team to review a PR. If `login` is provided, it does nothing if that user is already part of the team
   * You can also pass a `slack_webhook_url` to notify the assignees that they are assigned to the PR!
+### [**create-project-card**](.github/workflows/create-project-card.yml)
+  * It creates a card into your GitHub Project repository by providing a `project_name` and `project_destination_column_name` in which the card should be created.
```
danadajian, line 17:
```suggestion
  * Creates a Project card into your GitHub repository by providing a `project_name` and `project_destination_column_name` in which the card should be created.
```

README.md
```diff
@@ -13,6 +13,13 @@ Each of the following helpers are defined in a file of the same name in `src/hel
...
+  * If `note` is provided, it will add that information into the card. If it is not provided, it will use the PR information details to populate it.
+  * Useful when openning a pull request and want to track is information details into a GitHub Project.
```
danadajian, line 19:
```suggestion
  * Useful when opening a pull request and want to track is information details into a GitHub Project.
```

README.md
```diff
@@ -13,6 +13,13 @@ Each of the following helpers are defined in a file of the same name in `src/hel
...
+### [**move-project-card**](.github/workflows/move-project-card.yml)
+  * It moves a GitHub Project card from a `project_origin_column_name` to `project_destination_column_name` you provide.
```
danadajian, line 21:
```suggestion
  * Moves a GitHub Project card to a new column, using the `project_origin_column_name` and `project_destination_column_name` you provide.
```

README.md
```diff
@@ -13,6 +13,13 @@ Each of the following helpers are defined in a file of the same name in `src/hel
...
+  * In order to move a card from one place to another, it should exist first.
```
danadajian, line 22:
```suggestion
  * In order to move a card from one place to another, it must already exist.
```

src/helpers/create-project-card.ts
```diff
@@ -0,0 +1,64 @@
...
+
+  if (destinationColumn) {
+    return octokit.projects.createCard(cardParams);
+  } else {
+    core.info('No destination column was found');
+    return;
+  }
+};
+
+const generateCardParams = (filteredColumn: any, pullRequest: PullRequest, note?: string) => {
```
danadajian, line 48: Can we add a type for `filteredColumn`?

danadajian (approved):

PR #34, opened 2021-12-10
fix: check-pr-title regex to include "BREAKING CHANGE" and add to PR checks

aschwenn:
[to be more consistent with semantic release standards](https://github.com/semantic-release/semantic-release)

wasn't sure if this should trigger on `edited` as well, but people can always rerun the check fairly easily


.github/workflows/tests.yml
```diff
@@ -44,3 +44,12 @@ jobs:
...
           path: dist/
+  title:
+    name: Check PR title
+    runs-on: ubuntu-latest
+    steps:
+      - uses: ExpediaGroup/github-helpers@v1
+        with:
+          helper: check-pr-title
+          pull_number: ${{ github.event.pull_request.number }}
+          github_token: ${{ secrets.GITHUB_TOKEN }}
```
danadajian, line 55: We already have this workflow setup in the repo actually

aschwenn: oh lol i hadn't pulled the latest from `main`, i'll remove this

danadajian (approved):

PR #41, opened 2021-12-16
feat: add create-pr helper

wbautistaEXPEDIA:
### :pencil: Description
Implemented create pull request helper. Added tests. Updated documentation. Added workflow.


.github/workflows/create-pr.yml
```diff
@@ -0,0 +1,21 @@
...
+      - name: Checkout
+        uses: actions/checkout@v2
+
+      - uses: ./
+        with:
+          helper: create-pr
+          title: New PR
+          target: main
+          body: Implemented new feature. Added tests.
+          github_token: ${{ secrets.GITHUB_TOKEN }}
```
danadajian, line 21: This will likely require more than just the default github token. I think you will need a personal access token, similar to other helpers

src/helpers/create-pr.ts
```diff
@@ -0,0 +1,39 @@
...
+  source?: string;
+  target: string;
+  body: string;
+  modify?: boolean;
+  draft?: boolean;
+  issue?: string;
+}
+
+export const createPr = ({ title, source, target, body, modify, draft, issue }: CreatePR) => {
+  const sourceToUse = source ? source : context.ref.replace('refs/heads/', '');
```
danadajian, line 28:
```suggestion
  const sourceToUse = source || context.ref.replace('refs/heads/', '');
```

src/helpers/create-pr.ts
```diff
@@ -0,0 +1,39 @@
...
+import { octokit } from '../octokit';
+
+interface CreatePR {
+  title: string;
+  source?: string;
+  target: string;
+  body: string;
+  modify?: boolean;
+  draft?: boolean;
+  issue?: string;
```
danadajian, line 24: You will need to add any new inputs to `action.yml`

src/helpers/create-pr.ts
```diff
@@ -0,0 +1,39 @@
...
+export const createPr = ({ title, source, target, body, modify, draft, issue }: CreatePR) => {
+  const sourceToUse = source ? source : context.ref.replace('refs/heads/', '');
+  octokit.pulls.create({
+    title,
+    head: String(sourceToUse),
+    base: String(target),
+    body,
+    maintainer_can_modify: modify,
+    draft,
+    issue: issue ? Number(issue) : undefined,
```
danadajian, line 36: Why is "issue" needed?

aschwenn: i think if you want to tie it to an issue number, looks like it's optional tho

src/helpers/create-pr.ts
aschwenn: [alternatively](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_operator)

danadajian: I considered `??` but undefined inputs to these actions will be `''` which wouldn't satisfy that operator

src/helpers/create-pr.ts
```diff
@@ -0,0 +1,39 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+import { context } from '@actions/github';
+import { octokit } from '../octokit';
+
+interface CreatePR {
+  title: string;
+  source?: string;
```
danadajian, line 19:
```suggestion
  source_branch?: string;
```

src/helpers/create-pr.ts
```diff
@@ -0,0 +1,39 @@
...
+  target: string;
```
danadajian, line 20:
```suggestion
  target_branch: string;
```

src/helpers/create-pr.ts
danadajian, line 24:
```suggestion
  issue_number?: string;
```

src/helpers/create-pr.ts
```diff
@@ -0,0 +1,39 @@
...
+  draft?: boolean;
+  issue?: string;
+}
+
+export const createPr = ({ title, source, target, body, modify, draft, issue }: CreatePR) => {
+  const sourceToUse = source ? source : context.ref.replace('refs/heads/', '');
+  octokit.pulls.create({
+    title,
+    head: String(sourceToUse),
+    base: String(target),
```
danadajian, line 32: These are already strings, no need to stringify

.github/workflows/create-pr.yml
```diff
@@ -0,0 +1,21 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+
+      - uses: ./
+        with:
+          helper: create-pr
+          title: New PR
+          target: main
```
danadajian, line 19: We can default this value to `main` in action.yml

README.md
```diff
@@ -64,6 +64,8 @@ Additionally, the following parameters can be used for additional control over t
   * Updates a Github [deployment status](https://docs.github.com/en/rest/reference/repos#deployments)
 ### [**set-latest-pipeline-status**](.github/workflows/set-latest-pipeline-status.yml)
   * Determines whether the pipeline is clear for a PR. This means it will set the "pipeline" commit status to `pending` if there is an in-progress production deployment for the repo, and `success` otherwise.
+### [**create-pr**](.github/workflows/create-pr.yml)
+  * Opens a pull request
```
danadajian, line 68: Could we alphabetize this?

src/helpers/create-pr.ts
danadajian: Would we ever not want to create a PR with target = `default_branch`?

src/helpers/create-pr.ts
danadajian, line 36: I would kind of rather leave these as defaults if they aren't going to be used/needed in the real world.

src/helpers/create-pr.ts
wbautistaEXPEDIA: I imagine there could be some use cases, such as forks(?) and the desire to create PRs against a release branch, if a particular repo uses that branching strategy.

.github/workflows/create-pr.yml
wbautistaEXPEDIA: I don't see any other default values in `action.yml`. What makes this the first candidate, and what is the format?

.github/workflows/create-pr.yml
wbautistaEXPEDIA: Updated.

src/helpers/create-pr.ts
wbautistaEXPEDIA: Addressed. Thank you.

src/helpers/create-pr.ts
wbautistaEXPEDIA: Done.

src/helpers/create-pr.ts
wbautistaEXPEDIA: Done.

src/helpers/create-pr.ts
wbautistaEXPEDIA: Implemented the suggested change.

src/helpers/create-pr.ts
wbautistaEXPEDIA: Done.

src/helpers/create-pr.ts
wbautistaEXPEDIA: 🤦🏽 
Updated, thanks.

README.md
wbautistaEXPEDIA: Done.

src/helpers/create-pr.ts
```diff
@@ -0,0 +1,39 @@
...
+  draft?: boolean;
+  issue_number?: string;
+}
+
+export const createPr = ({ title, source_branch, target_branch, body, modify, draft, issue_number }: CreatePR) => {
+  const sourceToUse = source_branch || context.ref.replace('refs/heads/', '');
+  octokit.pulls.create({
+    title,
+    head: sourceToUse,
+    base: target_branch,
```
danadajian, line 32: Can we use `HEAD` here instead of requiring a `target_branch` input?

src/helpers/create-pr.ts
```diff
@@ -0,0 +1,39 @@
...
+    body,
+    maintainer_can_modify: modify,
+    draft,
+    issue: issue_number ? Number(issue_number) : undefined,
```
danadajian, line 36: Is passing `issue` necessary? What benefit does this serve? Could the issue number not simply be included in the PR description?

src/helpers/create-pr.ts
```diff
@@ -0,0 +1,39 @@
...
+
+export const createPr = ({ title, source_branch, target_branch, body, modify, draft, issue_number }: CreatePR) => {
+  const sourceToUse = source_branch || context.ref.replace('refs/heads/', '');
+  octokit.pulls.create({
+    title,
+    head: sourceToUse,
+    base: target_branch,
+    body,
+    maintainer_can_modify: modify,
+    draft,
```
danadajian, line 35: Are there use cases for creating draft PRs?

src/helpers/create-pr.ts
```diff
@@ -0,0 +1,39 @@
...
+}
+
+export const createPr = ({ title, source_branch, target_branch, body, modify, draft, issue_number }: CreatePR) => {
+  const sourceToUse = source_branch || context.ref.replace('refs/heads/', '');
+  octokit.pulls.create({
+    title,
+    head: sourceToUse,
+    base: target_branch,
+    body,
+    maintainer_can_modify: modify,
```
danadajian, line 34: Can we pick the value that serves the most common use case?

src/helpers/create-pr.ts
```diff
@@ -0,0 +1,39 @@
...
+  source_branch?: string;
+  target_branch: string;
+  body: string;
+  modify?: boolean;
+  draft?: boolean;
+  issue_number?: string;
+}
+
+export const createPr = ({ title, source_branch, target_branch, body, modify, draft, issue_number }: CreatePR) => {
+  const sourceToUse = source_branch || context.ref.replace('refs/heads/', '');
```
danadajian, line 28: Will this helper ever be used outside of a pull request workflow? If not I don't think the `source_branch` input is necessary

.github/workflows/create-pr.yml
danadajian: Sorry, I commented below about the `target_branch` input. It may not be needed

danadajian: Also, the idea behind these workflows is that it provides an opportunity to test your helper on this repo!

src/helpers/create-pr.ts
wbautistaEXPEDIA: Done.

src/helpers/create-pr.ts
wbautistaEXPEDIA: Not necessary. Removed.

src/helpers/create-pr.ts
wbautistaEXPEDIA: I don't have one. Defaulted this value to `false`.

src/helpers/create-pr.ts
wbautistaEXPEDIA: Done.

src/helpers/create-pr.ts
wbautistaEXPEDIA: Removed, and now using that default (latter of the logic above).

src/helpers/create-pr.ts
```diff
@@ -0,0 +1,33 @@
...
+
+export const createPr = ({ title, body }: CreatePR) => {
+  octokit.pulls.create({
+    title,
+    head: context.ref.replace('refs/heads/', ''),
+    base: 'HEAD',
+    body,
+    maintainer_can_modify: true,
+    draft: false,
+    issue: undefined,
```
danadajian, line 30: Nit pick, but we can omit these since they are not required

danadajian (approved):

src/helpers/create-pr.ts
wbautistaEXPEDIA: Done.

src/helpers/create-pr.ts
```diff
@@ -0,0 +1,31 @@
...
+export const createPr = ({ title, body }: CreatePR) => {
+  octokit.pulls.create({
+    title,
+    head: context.ref.replace('refs/heads/', ''),
+    base: 'HEAD',
+    body,
+    maintainer_can_modify: true,
+    ...context.repo
+  });
+};
```
danadajian, line 31:
```suggestion
export const createPr = ({ title, body }: CreatePR) => {
  const { data: { default_branch }} = await octokit.repos.get({...context.repo});
  return octokit.pulls.create({
    title,
    head: context.ref.replace('refs/heads/', ''),
    base: default_branch,
    body,
    maintainer_can_modify: true,
    ...context.repo
  });
};
```

PR #45, opened 2021-12-26
refactor: approve pr helper

danadajian:
### :pencil: Description
Refactors/renames the `auto-approve-pr` helper to be a simpler and generic `approve-pr` helper since the same "auto approve" workflow can be accomplished.


action.yml
```diff
@@ -79,9 +79,6 @@ inputs:
   paths_no_filter:
     description: 'A list of file paths to be included regardless of the filter (newline/comma separated)'
     required: false
-  auto_approved_user:
```
aschwenn, line 82: without digging through the codebase, i'm guessing the `login` param is used elsewhere and doesn't need to be removed?

aschwenn (approved):

action.yml
danadajian: Yep exactly! Also going to wait until I coordinate with someone who is the only known user of this helper since it's a breaking change and I'm cheating by not releasing a v2 for it 😁

PR #58, opened 2022-01-05
fix: tries passing the single required header to the projects api

salvadorgonmo:


danadajian:
I think this is the header we need!
```
application/vnd.github.starfox-preview+json
```

salvadorgonmo:
Closing it in favor of https://github.com/ExpediaGroup/github-helpers/pull/59

PR #59, opened 2022-01-05
fix: adds missing preview header for projects api

salvadorgonmo:
### :pencil: Description
Adds missing header for projects API. It is needed cause it is still in preview mode.

https://docs.github.com/en/graphql/overview/schema-previews#project-event-details-preview


src/constants.ts
```diff
@@ -14,7 +14,8 @@ limitations under the License.
 // These extra headers are for experimental operations. Newer versions of octokit may not require this
 export const GITHUB_OPTIONS = {
   headers: {
-    accept: 'application/vnd.github.ant-man-preview+json,application/vnd.github.flash-preview+json,application/vnd.github.v3+json'
+    accept:
+      'application/vnd.github.ant-man-preview+json,application/vnd.github.flash-preview+json,application/vnd.github.v3+json,application/vnd.github.starfox-preview+json'
```
danadajian, line 18: Could we delete the `application/vnd.github.v3+json`?

salvadorgonmo: Sure, removed in https://github.com/ExpediaGroup/github-helpers/pull/59/commits/150313942d9eb1172c088c6845800ea2b24f3d09#diff-8fa4b52909f895e8cda060d2035234e0a42ca2c7d3f8f8de1b35a056537bf199R18

danadajian (approved):

PR #106, opened 2022-01-28
fix(initiate-deployment): add auto_merge parameter

pongraczl:
### :pencil: Description
Set `auto_merge` parameter in `initiate-deployment` helper always to `false`.
- Parameter description from [Github docs](https://docs.github.com/en/rest/reference/deployments#create-a-deployment--parameters): _"Attempts to automatically merge the default branch into the requested ref, if it's behind the default branch."_ (the original, default value was `auto_merge: true`)
- ~~The default value is true (this is what initiate-deployment is currently using).~~
- ~~Adding this parameter gives more control over the behavior of this helper.~~

### :link: Related Issues
When a PR is not up-to-date with master, `auto_merge=true` makes the helper failing, but being up-to-date with master is not a requirement in all cases.


danadajian:
@pongraczl Thanks for finding this! Instead of adding an additional input, could we instead always set `auto_merge: false` for the helper?

pongraczl:
@danadajian Thanks! Then I simplified those changes to fixed value with `auto_merge: false`.
(Previously I was not sure if the original true value should remain available to some users.)

danadajian (approved):

PR #110, opened 2022-02-04
fix: ignore empty string in context of set-commit-status

aschwenn:
### :pencil: Description
i have a use case that looks like:
```
- name: Set PR check statuses if applicable
    uses: actions/github-helpers@v1
    with:
      helper: set-commit-status
      sha: ${{ github.event.pull_request.head.sha }}
      github_token: ${{ secrets.GITHUB_TOKEN }}
      context: |
        ${{ steps.paths.outputs.output != 'true' && 'build' || '' }}
        ${{ steps.paths.outputs.output != 'true' && 'unit-tests' || '' }}
        ...
      state: success
      description: This test can be skipped for this pull request
```
and I would like to avoid sending an API request when an empty string. It does nothing, but it potentially wastes an API hit, which is a (minor) problem for our repo.


src/helpers/set-commit-status.ts
```diff
@@ -25,13 +25,15 @@ export class SetCommitStatus {
...
-    octokit.repos.createCommitStatus({
-      sha,
-      context,
-      state: state as PipelineState,
-      description,
-      target_url,
-      ...githubContext.repo
-    })
+  map(
+    context.split('\n').filter(str => str !== ''),
```
danadajian, line 29: I think this should work fine and looks a bit nicer. If not lmk
```suggestion
    context.split('\n').filter(Boolean),
```

danadajian (approved):

PR #123, opened 2022-02-15
feat: new helper to re-run pr workflows

spincsak:
### :pencil: Description
Working on a new helper to re-run all PR workflows.


.github/workflows/rerun-pr-workflows.yml
```diff
@@ -0,0 +1,23 @@
+name: Rerun Pr Workflows
+
+on:
+  pull_request_target:
```
aschwenn, line 4: we can probably just switch this to utilize an example trigger, like on a comment add with a certain substring, or a specific label add, rather than on `pull_request_target`

README.md
```diff
@@ -40,6 +40,9 @@ with:
 
 ## Available Helpers
 Each of the following helpers are defined in a file of the same name in `src/helpers`:
+### [rerun-pr-workflows](.github/workflows/rerun-pr-workflows.yml)
+  * use to rerun all failed workflows on a pull request
```
aschwenn, line 44: hmm this is a good point.. should we run _all_ workflows or just _failed_ workflows? i would imagine there's no need to rerun successful workflows unless they somehow failed into a success because of something like rate limiting lol. unless filtering in this way represents a development hurdle

src/helpers/rerun-pr-workflows.ts
```diff
@@ -0,0 +1,41 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+import { context as githubContext } from '@actions/github';
+import { octokit } from '../octokit';
+
+export class RerunPrWorkflows {
+  branch = '';
+  //maybe fork name should be an optional input?
```
aschwenn, line 19: i'm thinking that maybe if we keep the trigger on `pull_request` instead of `pull_request_target` it should automatically get the fork from the context. we likely couldn't force reruns on a fork from `pull_request_target` anyway as the tokens/PATs wouldn't have scope into the fork repo (i'm guessing)

src/helpers/rerun-pr-workflows.ts
```diff
@@ -0,0 +1,41 @@
...
+  // forkName?: string;
+}
+
+export const rerunPrWorkflows = async ({ branch }: RerunPrWorkflows) => {
+  // list all workflow runs
+  const allRuns = await octokit.request('GET /repos/{owner}/{repo}/actions/{branch}', {
```
aschwenn, line 25: didn't know you could do this haha. i think to be in line with the other helpers we may want to use [this method](https://octokit.github.io/rest.js/v18#actions-list-workflow-runs) which accomplishes the same thing

src/helpers/rerun-pr-workflows.ts
```diff
@@ -0,0 +1,41 @@
...
+    repo: githubContext.repo.repo,
+    branch
+  });
+  // store the latest run of each workflow
+  //double-check the structure of the response.
+  const latest = allRuns.data.workflow_runs[0];
+
+  // each workflow run instance has a rerun_url
+  // trigger a POST request to the rerun_url to force a rerun
+  await octokit.request('POST /repos/{owner}/{repo}/{branch}/actions/runs/{run_id}/rerun', {
```
aschwenn, line 36: probably here too with [this](https://octokit.github.io/rest.js/v18#actions-get-workflow-run)

src/helpers/rerun-pr-workflows.ts
```diff
@@ -0,0 +1,41 @@
...
+export const rerunPrWorkflows = async ({ branch }: RerunPrWorkflows) => {
+  // list all workflow runs
+  const allRuns = await octokit.request('GET /repos/{owner}/{repo}/actions/{branch}', {
+    owner: githubContext.repo.owner,
+    repo: githubContext.repo.repo,
+    branch
+  });
+  // store the latest run of each workflow
+  //double-check the structure of the response.
+  const latest = allRuns.data.workflow_runs[0];
```
aschwenn, line 32: probably need to iterate here to find the latest run of each test triggered by the most recent commit

README.md
spincsak: Based on the comments in the [stack overflow thread](https://stackoverflow.com/questions/56435547/how-do-i-re-run-github-actions/59939257#59939257) I've been referencing, it looks like successful workflows can't be re-run. 

(I haven't tried it yet, though, can't 100% confirm that it's true.)

src/helpers/rerun-pr-workflows.ts
spincsak: I don't know if my method actually works, because it is failing on that step when I try to run the unit tests. I'll refactor to use the other method and see how that goes.

README.md
aschwenn: ah you may be right. guess that solves it for us 🤷‍♂️

src/helpers/rerun-pr-workflows.ts
aschwenn: (as long as this still gives us the same data)

danadajian: Yeah let's be consistent and use the [octokit](https://github.com/ExpediaGroup/github-helpers/blob/main/src/octokit.ts#L18) object for github API calls

PR #139, opened 2022-03-04
feat(generate-path-matrix): allow globs to be used instead of paths

aschwenn:
### :pencil: Description
An internal repo has a more specific use case for this, so rather than all package data being contained in `package/package-name/*`, it may exist in `package/package-name`, `test/package-name`, `docs/package-name`, etc. This change should help allow them to use this helper for their setup.


src/helpers/generate-path-matrix.ts
```diff
@@ -11,34 +11,43 @@ See the License for the specific language governing permissions and
...
+  globs = '',
+  /** paths that override the changed files filter, causing the action to return all paths */
   override_filter_paths,
   override_filter_globs,
+  /** paths that will be returned regardless of their adherence to the filter */
   paths_no_filter,
+  /** number of evenly-sized batches to separate matching paths into (returns comma-separated result) */
   batches
 }: GeneratePathMatrix) => {
+  if (!paths && !globs) core.error('Must supply one of paths, globs');
```
danadajian, line 40: Let's `throw new Error()` here too

src/helpers/generate-path-matrix.ts
```diff
@@ -11,34 +11,43 @@ See the License for the specific language governing permissions and
...
+  globs?: string;
   override_filter_paths?: string;
   override_filter_globs?: string;
   paths_no_filter?: string;
   batches?: string;
 }
 
 export const generatePathMatrix = async ({
-  paths,
+  paths = '',
```
danadajian, line 30: Inputs are already defaulted to empty string

danadajian (approved):

src/helpers/generate-path-matrix.ts
aschwenn: i changed the interface definition to set `paths`/`globs` as optional since they have an XOR relationship, so i moved their initialization to null string here

danadajian: I'm just saying that `paths = ''` is actually redundant here

danadajian: The `getInput` method in `@actions/core` returns empty string if no value is passed to the action, and that's what we [use](https://github.com/ExpediaGroup/github-helpers/blob/b7e1485b7ba1a900c5dbdc00aa47bffb2649d050/src/utils/get-action-inputs.ts#L25) to read in all inputs dynamically

danadajian: But anyway, I still don't think we should default `paths` or `globs` to empty string to stay consistent with the rest. We should just handle the undefined case properly in the helper I think

src/helpers/generate-path-matrix.ts
```diff
@@ -11,34 +11,47 @@ See the License for the specific language governing permissions and
...
   override_filter_globs,
+  /** paths that will be returned regardless of their adherence to the filter */
   paths_no_filter,
+  /** number of evenly-sized batches to separate matching paths into (returns comma-separated result) */
   batches
 }: GeneratePathMatrix) => {
+  if (!paths && !globs) {
+    const err = 'Must supply one of paths, globs';
+    core.error(err);
+    throw new Error(err);
```
danadajian, line 43: Nit pick, but probably can just do this so the error doesn't log twice, and to be consistent with other errors in this repo

```suggestion
    core.error('Must supply one of paths, globs');
    throw new Error();
```

danadajian:
@aschwenn Made some tweaks, lemme know if you think they make sense!

aschwenn:
looks good, i'm gonna merge

Issue #153, opened 2022-03-16
Intermittent Malformed API Response in GHE Server 3.2.10


danadajian:
**Describe the bug**

On GitHub Enterprise Server v3.2.10, there is an issue with the commit status endpoint in API v3 which is resulting in some job failures with the following error:

```
Error: invalid json response body at https:/GITHUB_URL/api/v3/repos/org/repo/statuses/eb1906c4ef99838204c8169c348ddb2c993d228a reason: Unexpected end of JSON input
```

**To Reproduce**

Only happens sporadically and not predictably reproducible.

danadajian:
It appears this is in fact an issue with GHE Server and not github-helpers.

PR #155, opened 2022-03-17
fix: add retry functionality to octokit fetch

danadajian:
### :pencil: Description
The octokit library supports passing in a [custom fetch method](https://octokit.github.io/types.ts/modules/RequestRequestOptions.html), which we can use to pass in a fetch method with built-in retryability!

### :link: Related Issues
#153


src/octokit.ts
```diff
@@ -12,8 +12,9 @@ limitations under the License.
 */
 
 import * as core from '@actions/core';
+import * as fetch from '@adobe/node-fetch-retry';
 import { Octokit } from '@octokit/rest';
 import { getOctokit } from '@actions/github';
 
 const githubToken = core.getInput('github_token', { required: true });
-export const octokit = getOctokit(githubToken, { userAgent: 'github-helpers' }).rest as unknown as Octokit;
```
aschwenn, line 19: i'm guessing we don't need `userAgent` anymore?

aschwenn (approved):
looks cool 🚀

src/octokit.ts
danadajian: I think octokit adds a default user agent anyways

PR #156, opened 2022-03-18
feat: allow manage-merge-queue to paginate all PRs

aschwenn:
The `manage-merge-queue` helper will now list _all_ PRs in a repo when determining the merge queue.


danadajian (commented):
We could just use `Bluebird.map()` to map over the array `[1,2]`, make the call with those page values, and aggregate the results?

src/helpers/manage-merge-queue.ts
```diff
@@ -82,12 +82,23 @@ const addPrToQueue = async (pullRequest: PullRequest, queuePosition: number) =>
...
+      sort: 'updated',
+      direction: 'desc',
+      per_page: 100,
+      page,
+      ...context.repo
+    });
+    // eslint-disable-next-line functional/immutable-data
+    openPullRequests.push(...response.data);
+    page += 1;
+  } while (response.data.length);
```
danadajian, line 102: Could we do something similar to [this](https://github.com/ExpediaGroup/github-helpers/blob/main/src/helpers/rerun-pr-checks.ts#L35-L44)? That way we don't have to ignore all the lint rules 😄

danadajian: And an example of how to test it is [here](https://github.com/ExpediaGroup/github-helpers/blob/main/test/helpers/rerun-pr-checks.test.ts)

danadajian: Also, do we actually need info about over 100 PRs if the merge queue is less than 100 PRs long?

aschwenn: that still wouldn't allow us to grab all of the open pulls. correct me if i'm wrong, but this logic isn't about grabbing just the pulls in the merge queue, but grabbing all of them and then filtering just the ones with the appropriate tag since github doesn't allow us to do this through the API. so if we have >100 open PRs and don't do this, we could potentially be missing PRs with the `READY FOR MERGE` tag when constructing the merge queue because they aren't in the latest 100 updated PRs

danadajian: I see. My original thought was that since we sort PRs by last updated we would account for all PRs with merge queue activity, but that may not be the case I guess if there's tons of activity in a repo.

danadajian (approved):

src/helpers/manage-merge-queue.ts
aschwenn: yeah it's more of an edge case thing, just looking to stop this before we get to where it might become a problem someone has to debug

PR #159, opened 2022-03-23
feat: rerun-pr-checks

aschwenn:


src/helpers/rerun-pr-checks.ts
```diff
@@ -0,0 +1,55 @@
...
+    return;
+  }
+  /** grab only latest occurrence of each workflow run */
+  const latestHash = workflowRuns.data.workflow_runs[0].head_sha;
+  const latestRuns = workflowRuns.data.workflow_runs.filter(({ head_sha }) => head_sha === latestHash);
+  core.info(`Found the ${latestRuns} latest runs on this branch, triggering reruns...`);
+  /** trigger a rerun for all of the latest runs on the branch */
+  latestRuns.forEach(async ({ id, name }) => {
+    core.info(`- Rerunning ${name}`);
+    await octokit.request('POST /repos/{owner}/{repo}/actions/runs/{run_id}/rerun', {
```
danadajian, line 49: Guessing there's no support for this endpoint in octokit?

src/helpers/rerun-pr-checks.ts
```diff
@@ -0,0 +1,55 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+import * as core from '@actions/core';
+import { context } from '@actions/github';
+import { octokit } from '../octokit';
+
+export const rerunPrChecks = async () => {
+  /** grab owner in case of fork branch */
```
danadajian, line 19: Why do we need to do this here and not in any other helper?

test/helpers/rerun-pr-checks.test.ts
```diff
@@ -0,0 +1,106 @@
...
+
+    expect(octokit.actions.listWorkflowRunsForRepo).toHaveBeenCalledWith({
+      branch: 'branch',
+      repo: 'repo',
+      owner: 'owner',
+      event: 'pull_request',
+      per_page: 100
+    });
+
+    expect(octokit.request).toHaveBeenCalledTimes(3);
```
danadajian, line 83: Could we assert on the id's this is called with to prove the filtering part works?

src/helpers/rerun-pr-checks.ts
```diff
@@ -0,0 +1,55 @@
...
+    event: 'pull_request',
+    per_page: 100
+  });
+  if (!workflowRuns.data.total_count) {
+    core.info(`No workflow runs found on branch ${branch} on ${owner}/${context.repo.repo}`);
+    return;
+  }
+  /** grab only latest occurrence of each workflow run */
+  const latestHash = workflowRuns.data.workflow_runs[0].head_sha;
+  const latestRuns = workflowRuns.data.workflow_runs.filter(({ head_sha }) => head_sha === latestHash);
```
danadajian, line 44: A bit confused by this part - if you update the test to assert which ids are passed in I'll take another look. Also I slightly prefer `.find(Boolean)` to `[0]`, but if you wanna go with `[0]` let's make it `[0]?.head_sha`

src/helpers/rerun-pr-checks.ts
aschwenn: since there's no way to grab workflow runs on a PR at the moment, i'm filtering workflow runs by branch and grabbing only the first occurrence. if the `head` is on a fork, i need to query that repo instead. this sort of hinges on the assumption that the forked repo is publicly available

test/helpers/rerun-pr-checks.test.ts
aschwenn: sure

src/helpers/rerun-pr-checks.ts
aschwenn: `[0]` _should_ always be accessible given that the previous condition returns if `total_count === 0`. but i'll add `?.` for safety just in case GitHub goofs in its response somehow

aschwenn: actually, to make this clearer, i think a more elegant way to grab the latest commit hash would just be to do so from the metadata returned from `pulls.get()`. i'll update it to reflect that

spincsak:
Thanks for sharing this with me, Andrew! It's really helpful to compare the way you've executed this with my previous attempt.

danadajian:
@aschwenn Should we be concerned that it doesn't work on this PR? 
```
No workflow runs found on branch refs/pull/159/merge on ExpediaGroup/github-helpers
```

aschwenn:
> @aschwenn Should we be concerned that it doesn't work on this PR?
> 
> ```
> No workflow runs found on branch refs/pull/159/merge on ExpediaGroup/github-helpers
> ```

oops yeah i was still testing it

danadajian:
@aschwenn I got it working by pulling in the `@octokit/request` library. However, it only re-runs checks that have the `pull_request` trigger, not the `pull_request_target` trigger.

danadajian (approved):
Got things working, this will now re-run `pull_request` and `pull_request_target` workflows! If the workflow is in progress it'll just log that's the case.

aschwenn:
awesome, appreciate the help!

PR #160, opened 2022-03-24
feat(create-pr): add a new base parameter

JordanJLopez:
### :pencil: Description

The ability to create pull requests with actions is very helpful, but prior to this PR the only option was to create pull requests onto the default branch. Some teams, my own included, maintain several different branches that would benefit from automatic PR creation.

This PR adds the ability to specify a base for the newly created pull requests.


src/helpers/create-pr.ts
```diff
@@ -18,20 +18,26 @@ export class CreatePR {
...
   const pullNumber = result?.data?.number;
   return pullNumber;
 };
+
+async function getDefaultBranch() {
+  const {
+    data: { default_branch }
+  } = await octokit.repos.get({ ...context.repo });
+  return default_branch;
+}
```
JordanJLopez, line 43: By refactoring this into a separate method, I was able to use the elvis operator to prevent unnecessary API calls.

src/helpers/create-pr.ts
```diff
@@ -18,20 +18,26 @@ export class CreatePR {
...
   head?: string;
+  base?: string;
 }
 
-export const createPr = async ({ title, body, head = context.ref.replace('refs/heads/', '') }: CreatePR) => {
-  const {
-    data: { default_branch }
-  } = await octokit.repos.get({ ...context.repo });
+export const createPr = async ({ title, body, head = context.ref.replace('refs/heads/', ''), base }: CreatePR) => {
+  const pr_base = base != null ? base : await getDefaultBranch();
```
danadajian, line 25: `base` will be `''` if omitted

```suggestion
  const pr_base = base || await getDefaultBranch();
```

danadajian (approved):

PR #167, opened 2022-03-28
chore: add debug statement to error handler when updating branch in pr queue

aschwenn:
having some issues with the error getting swallowed here, so logging it with the debugger so that enabling debug mode will show the full error


src/helpers/prepare-queued-pr-for-merge.ts
```diff
@@ -46,9 +46,10 @@ export const updatePrWithMainline = async (pullRequest: PullRequest | SimplePull
   } catch (error) {
     if ((error as GithubError).status === 204) {
       core.info('The first PR in the queue is already up to date!');
-    }
-    if ((error as GithubError).status === 409) {
+    } else if ((error as GithubError).status === 409) {
       core.info('The first PR in the queue has a merge conflict.');
+    } else {
+      core.debug((error as GithubError).message);
```
danadajian, line 52:
```suggestion
      core.info((error as GithubError).message);
```

src/helpers/prepare-queued-pr-for-merge.ts
```diff
@@ -46,9 +46,9 @@ export const updatePrWithMainline = async (pullRequest: PullRequest | SimplePull
   } catch (error) {
     if ((error as GithubError).status === 204) {
       core.info('The first PR in the queue is already up to date!');
-    }
-    if ((error as GithubError).status === 409) {
-      core.info('The first PR in the queue has a merge conflict.');
+    } else if ((error as GithubError).status >= 400) {
+      if ((error as GithubError).status === 409) core.error('The first PR in the queue has a merge conflict.');
+      else core.error((error as GithubError).message);
```
danadajian, line 51: So we already have a generic error handler [here](https://github.com/ExpediaGroup/github-helpers/blob/main/src/main.ts#L27). Will this do anything different?

src/helpers/prepare-queued-pr-for-merge.ts
```diff
@@ -46,9 +46,9 @@ export const updatePrWithMainline = async (pullRequest: PullRequest | SimplePull
   } catch (error) {
     if ((error as GithubError).status === 204) {
       core.info('The first PR in the queue is already up to date!');
-    }
-    if ((error as GithubError).status === 409) {
-      core.info('The first PR in the queue has a merge conflict.');
+    } else if ((error as GithubError).status >= 400) {
```
danadajian, line 49: Won't we want to log out an error with status < 400?

src/helpers/prepare-queued-pr-for-merge.ts
aschwenn: nah, should be interpreted in the same way internally. just saw we used `core.error` in some places in the code

src/helpers/prepare-queued-pr-for-merge.ts
aschwenn: anything in the 200s is still a success message so i don't really think it's worth failing the job over. i guess i'm assuming that 100s/300s won't ever happen with the way we're using octokit for all of our API requests

danadajian: I'm talking about logging with `core.info`. As it stands we wouldn't do anything with status < 400 right? Since if that if statement condition is not met it just skips that whole block?

src/helpers/prepare-queued-pr-for-merge.ts
danadajian: I believe we only use `core.error` when there is a specific error to log, not a generic one. Could be wrong tho!

src/helpers/prepare-queued-pr-for-merge.ts
aschwenn: ah yeah i can just log it otherwise, or just always log it and set failure based on these conditions

src/helpers/prepare-queued-pr-for-merge.ts
```diff
@@ -46,9 +46,11 @@ export const updatePrWithMainline = async (pullRequest: PullRequest | SimplePull
...
+      if ((error as GithubError).status === 409) core.setFailed('The first PR in the queue has a merge conflict.');
+      else core.setFailed((error as GithubError).message);
```
danadajian, line 51: Cool, makes sense. So now I dont think this line is necessary since it's duplicating the generic error handler in `main.ts`

danadajian (approved):

PR #171, opened 2022-03-30
fix: use rerun_url in rerun-pr-checks to ensure base url is correct

aschwenn:


src/helpers/rerun-pr-checks.ts
```diff
@@ -38,7 +38,8 @@ export const rerunPrChecks = async () => {
       ...context.repo,
       owner,
       event,
-      per_page: 100
+      per_page: 100,
+      status: 'completed'
```
aschwenn, line 42: this prevents it from trying to rerun itself...

danadajian (approved):

PR #228, opened 2022-06-30
feat(generate-path-matrix): allow for use of only paths_no_filter to avoid all filtering in generating matrix

aschwenn:
### :pencil: Description
- allows users to pass _only_ `paths_no_filter` to construct a matrix strategy regardless of changed files in PR

### :link: Related Issues
-


aschwenn:
unless you have other use cases for this functionality

PR #258, opened 2022-09-01
feat(manage-merge-queue): add configurability to auto-merge

emarler:
Adding the ability to have queue without auto-merge.

### :pencil: Description
- 

### :link: Related Issues
-

### DIFF
--- a/README.md
+++ b/README.md
@@ -113,7 +113,6 @@ Additionally, the following parameters can be used for additional control over t
  * You can also pass `login` and `slack_webhook_url` to notify the PR author when they are in the 1st position of the merge queue.
  * You can pass `auto_merge` as false to disable auto-merge for a PR when it is in the 1st position of the merge queue.
 
-
 ### [move-project-card](.github/workflows/move-project-card.yml)
 * Moves a GitHub Project card to a new column, using the `project_origin_column_name` and`project_destination_column_name` you provide.
 * In order to move a card from one place to another, it must already exist.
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -111,6 +111,7 @@ Additionally, the following parameters can be used for additional control over t
     * Adding the `JUMP THE QUEUE` label to a PR will make that PR first in the queue immediately.
     * When a PR is merged, it automatically updates the first-queued PR with the default branch.
  * You can also pass `login` and `slack_webhook_url` to notify the PR author when they are in the 1st position of the merge queue.
+ * You can pass `auto_merge` as false to disable auto-merge for a PR when it is in the 1st position of the merge queue.
```
aschwenn, line 114: might be more intuitive to do something like `no_auto_merge`, since i feel like the default behavior of any configurable boolean parameters should be false

emarler: Is this a breaking change for those that already use it?

src/helpers/manage-merge-queue.ts
```diff
@@ -92,6 +98,15 @@ const addPrToQueue = async (pullRequest: PullRequest, queuePosition: number) =>
     enableAutoMerge(pullRequest.node_id)
   ]);
 
+const addPrToQueue = async (queuePosition: number) =>
+  Promise.resolve(
+    octokit.issues.addLabels({
```
aschwenn, line 103: not a big deal but if this code block is duplicated it might make sense to abstract it, or rather just call it in `addPrToQueueWithAutoMerge`

emarler: Made update.

danadajian:
While I appreciate the contribution here, I don't see a use case for this. The `manage-merge-queue` helper is intended to mimic the [upcoming GitHub merge queue feature](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/configuring-pull-request-merges/managing-a-merge-queue), which is designed to automatically merge all PRs that enter the queue.

PR #265, opened 2022-09-15
fix(assign-pr-reviewers): remove author from core members

PoojaDonekal1294:
### :pencil: Description
-  To avoid assigning the author as a core approver

### :link: Related Issues
-


src/helpers/assign-pr-reviewers.ts
```diff
@@ -36,12 +36,19 @@ export const assignPrReviewers = async ({
...
   if (login && coreMemberLogins.includes(login)) {
     core.info('Already a core member, no need to assign.');
     return;
   }
-  const assignees = sampleSize(coreMemberLogins, Number(number_of_assignees));
+  let filteredCoreMemberLogins = coreMemberLogins;
+  if (user) {
+    filteredCoreMemberLogins = coreMemberLogins.filter(userName => userName != user.login);
+  }
+  const assignees = sampleSize(filteredCoreMemberLogins, Number(number_of_assignees));
```
danadajian, line 51:
```suggestion
  const prAuthorUsername = user?.login;
  const filteredCoreMemberLogins = coreMemberLogins.filter(userName => userName !== prAuthorUsername);
  const assignees = sampleSize(filteredCoreMemberLogins, Number(number_of_assignees));
```

danadajian: Even if `prAuthorUsername` is undefined (which I don't think will ever be the case 🤷 ), this filter should still work as expected since none of the `coreMemberLogins` will be undefined

danadajian (approved):

PR #298, opened 2022-11-29
fix(check-merge-safety): simplify check merge safety helper

danadajian:
### :pencil: Description
- Realized that the `paths` input needs to be the full list of a repo's project directories for the "push" workflow case
- Made stuff more readable
- Updated the tests to be more understandable (hopefully!)


src/helpers/check-merge-safety.ts
```diff
@@ -79,13 +82,14 @@ const getMergeSafetyMessage = async (
...
+    ? micromatch(fileNamesWhichBranchIsBehindOn, override_filter_globs.split('\n'))
     : override_filter_paths
-    ? fileNamesWhichBranchIsBehindOn.some(changedFile => override_filter_paths.split(/[\n,]/).includes(changedFile))
-    : false;
+    ? fileNamesWhichBranchIsBehindOn.filter(changedFile => override_filter_paths.split(/[\n,]/).includes(changedFile))
+    : [];
 
-  if (shouldOverrideSafetyCheck) {
+  if (globalFilesOutdatedOnBranch.length) {
+    core.error(buildErrorMessage(globalFilesOutdatedOnBranch));
```
aschwenn, line 92: throwing `the following projects are outdated` might be misleading for filenames like `package.json`

src/helpers/check-merge-safety.ts
```diff
@@ -96,12 +100,21 @@ const getMergeSafetyMessage = async (
     basehead: `${baseOwner}:${default_branch}...${username}:${ref}`
   });
   const changedFileNames = changedFiles?.map(file => file.filename);
-  const projectDirectories = paths?.split(/[\n,]/);
-  const isUnsafeToMerge = projectDirectories?.some(
+  const changedProjectDirectories = paths?.split(/[\n,]/);
```
aschwenn, line 103: are we expecting different usage from `paths` now?

src/helpers/check-merge-safety.ts
```diff
@@ -96,12 +100,21 @@ const getMergeSafetyMessage = async (
...
 
-  if (isUnsafeToMerge) {
+  if (changedProjectsOutdatedOnBranch?.length) {
+    core.error(buildErrorMessage(changedProjectsOutdatedOnBranch));
     return `This branch has one or more outdated projects. Please update with ${default_branch}.`;
   }
 };
+
+const buildErrorMessage = (paths: string[]) =>
+  `
```
aschwenn, line 116: nitpick but why does the error message start with a newline?

src/helpers/check-merge-safety.ts
danadajian: oh crap good catch, that was not intended!

src/helpers/check-merge-safety.ts
danadajian: No, this is just a refactor that allows us to log out the specific projects or files that caused the failure

src/helpers/check-merge-safety.ts
danadajian: Just looked a bit better to me in the GHA logs 🤷

src/helpers/check-merge-safety.ts
danadajian: `paths` should actually be the complete list of projects in the repo, not just the changed packages. While the list of changed packages would work for the PR workflow, it would not work for the push workflow since that workflow doesn't know the changed packages of each PR

aschwenn: so if `paths` is a complete list of projects, why is the list called `changedProjectDirectories`?

danadajian: That's a really good question, it shouldn't be. 🤦

aschwenn (approved):

PR #299, opened 2022-12-04
feat(get-changed-files): support filtering by regex input

tpoehlman:
### :pencil: Description
- Enhancing get-changed-files by adding a filter parameter
### :link: Related Issues
- [CSPB-549894]


danadajian (changes requested):
This should be an enhancement to the existing `get-changed-files` helper. Let's use the existing `glob_filter_paths` input and make it optional

test/helpers/get-changed-files.test.ts
```diff
@@ -24,7 +24,7 @@ jest.mock('@actions/github', () => ({
 const mock_data1 = [
   {
     sha: 'bbcd538c8e72b8c175046e27cc8f907076331401',
-    filename: 'file/path/1/file1.txt',
+    filename: 'src/main/resources/messages_en_US.properties',
```
danadajian, line 27: I don't think we need to change this mock file name

src/helpers/get-changed-files.ts
```diff
@@ -11,10 +11,16 @@ See the License for the specific language governing permissions and
...
-};
+export class checkChangedFiles extends HelperInputs {
+  glob_filter_paths?: string;
+}
+
+export const getChangedFiles = async ({ glob_filter_paths }: checkChangedFiles) =>
+  (await getChangedFilepaths(context.issue.number))
+    .map(fileName => fileName.match(glob_filter_paths || '[\\s\\S]*'))
+    ?.filter(localizationFile => localizationFile !== null)
+    .join(',');
```
danadajian, line 26: We only need to check regex if `glob_filter_paths` is provided. We should do something like this:
```suggestion
  const filePaths = await getChangedFilepaths(context.issue.number);
  const filteredFilePaths = glob_filter_paths ? filePaths.map(fileName => fileName.match(glob_filter_paths)) : filePaths;
  return filteredFilePaths.join(',');
```

danadajian (changes requested):

src/helpers/get-changed-files.ts
```diff
@@ -11,10 +11,18 @@ See the License for the specific language governing permissions and
...
-export const getChangedFiles = async () => {
+export class GetChangedFiles extends HelperInputs {
+  glob_filter_paths?: string;
+}
+
+export const getChangedFiles = async ({ glob_filter_paths }: GetChangedFiles) => {
   const filePaths = await getChangedFilepaths(context.issue.number);
-  return filePaths.join(',');
+  const filteredFilePaths = glob_filter_paths
+    ? filePaths.map(fileName => fileName.match(glob_filter_paths))?.filter(localizationFile => localizationFile !== null)
```
danadajian, line 25: Just realized you are looking for a regex pattern, not a glob pattern, right? There is an existing input called `pattern` for this
```suggestion
    ? filePaths.filter(fileName => fileName.test(pattern))
```

src/helpers/get-changed-files.ts
```diff
@@ -11,10 +11,18 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
+import { HelperInputs } from '../types/generated';
 import { context } from '@actions/github';
 import { getChangedFilepaths } from '../utils/get-changed-filepaths';
 
-export const getChangedFiles = async () => {
+export class GetChangedFiles extends HelperInputs {
+  glob_filter_paths?: string;
```
danadajian, line 19:
```suggestion
  pattern?: string;
```

src/helpers/get-changed-files.ts
```diff
@@ -11,10 +11,18 @@ See the License for the specific language governing permissions and
...
+}
+
+export const getChangedFiles = async ({ glob_filter_paths }: GetChangedFiles) => {
```
danadajian, line 22:
```suggestion
export const getChangedFiles = async ({ pattern }: GetChangedFiles) => {
```

test/helpers/get-changed-files.test.ts
```diff
@@ -67,8 +67,20 @@ const mock_data2 = [
 
 describe('getChangedFiles', () => {
   it('should return true if one of the file paths match the file paths that octokit returns', async () => {
-    const result = await getChangedFiles();
+    const result = await getChangedFiles({});
 
     expect(result).toEqual(`${mock_data1[0].filename},${mock_data1[1].filename},${mock_data2[0].filename}`);
   });
+
+  it('should return true if files returned from getChangedFiles contains the properties file', async () => {
```
danadajian, line 75:
```suggestion
  it('should return true if files returned from getChangedFiles matches the provided regex pattern', async () => {
```

test/helpers/get-changed-files.test.ts
```diff
@@ -67,8 +67,20 @@ const mock_data2 = [
...
+    const result = await getChangedFiles({ glob_filter_paths: 'file/path/1/file[0-9].txt' });
+
+    expect(result).toEqual(`${mock_data1[0].filename}`);
+  });
+
+  it('should return false if files returned from getChangedFiles do not contain the properties file', async () => {
```
danadajian, line 81:
```suggestion
  it('should return false if files returned from getChangedFiles does not match the provided regex pattern', async () => {
```

danadajian (changes requested):

README.md
```diff
@@ -139,6 +139,7 @@ Additionally, the following parameters can be used for additional control over t
 
 ### [get-changed-files](.github/workflows/get-changed-files.yml)
   * Returns a comma-separated list of changed files for a PR
+  * Include the regular expression parameter 'glob_filter_paths' to add a filter, it will return any filepath that matches.
```
danadajian, line 142:
```suggestion
  * Include the regular expression parameter 'pattern' to add a filter, it will return any filepath that matches.
```

PR #309, opened 2022-12-15
feat(approvals-satisfied): add approvals-satisfied helper

danadajian:
### :pencil: Description
- Adds the `approvals-satisfied` helper, which determines if a PR has been approved by each of the provided GitHub team(s) if provided or `CODEOWNERS` by default!


src/helpers/approvals-satisfied.ts
```diff
@@ -0,0 +1,28 @@
...
+
+export class ApprovalsSatisfied extends HelperInputs {
+  teams?: string;
+}
+
+export const approvalsSatisfied = async ({ teams }: ApprovalsSatisfied = {}) => {
+  const coreMemberLogins = await getCoreMemberLogins(context.issue.number, teams?.split('\n'));
+  const { data: reviews } = await octokit.pulls.listReviews({ pull_number: context.issue.number, ...context.repo });
+
+  return reviews.some(({ state, user }) => user?.login && coreMemberLogins.includes(user.login) && state === 'APPROVED');
```
aschwenn, line 27: will this work when multiple teams are requested for review and not all of them have yet approved?

danadajian: Ah yeah. Right now this creates an aggregated pool of all users on all codeowner teams (or all teams provided in the input). I will make it so it only returns true if at least 1 member of each team has approved the PR.

src/utils/get-core-member-logins.ts
```diff
@@ -13,37 +13,42 @@ limitations under the License.
...
   const codeOwners = teams ?? (await getCodeOwners(pull_number));
 
   if (!codeOwners?.length) {
-    core.setFailed('No code owners found.');
+    core.setFailed('No code owners found. Please provide a "teams" input or set up a CODEOWNERS file in your repo.');
     throw new Error();
   }
 
-  const adminLogins = await map(codeOwners, async team =>
+  const teamsAndLogins = await map(codeOwners, async team =>
```
aschwenn, line 35: totally out of curiosity.... does using the bluebird `map` function give us any advantage here over the array method?

src/helpers/approvals-satisfied.ts
```diff
@@ -0,0 +1,37 @@
...
+  teams?: string;
+}
+
+export const approvalsSatisfied = async ({ teams }: ApprovalsSatisfied = {}) => {
+  const { data: reviews } = await octokit.pulls.listReviews({ pull_number: context.issue.number, ...context.repo });
+  const teamsAndLogins = await getCoreTeamsAndLogins(context.issue.number, teams?.split('\n'));
+  const approverLogins = reviews
+    .filter(({ state }) => state === 'APPROVED')
+    .map(({ user }) => user?.login)
+    .filter((login): login is string => Boolean(login));
```
aschwenn, line 30: huh, never seen this syntax before

aschwenn (approved):

src/utils/get-core-member-logins.ts
danadajian: `bluebird` is specifically meant for asynchronous iteration. That said, I think it's an old library and we could probably replace it 🤷

src/helpers/approvals-satisfied.ts
danadajian: Me neither! I found it on stack overflow as a way to force Typescript to infer `approverLogins` as type `string[]`, since by default if you just do `.filter(Boolean)` it's too dumb for some reason and infers it as `(string | undefined)[]`. 🤷

PR #322, opened 2023-01-25
feat(set-commit-status): add skip_if_already_set param

aschwenn:
### :pencil: Description
- 

### :link: Related Issues
-


src/helpers/set-commit-status.ts
```diff
@@ -23,17 +24,33 @@ export class SetCommitStatus extends HelperInputs {
...
   description?: string;
   target_url?: string;
+  skip_if_already_set?: string;
 }
 
-export const setCommitStatus = async ({ sha, context, state, description, target_url }: SetCommitStatus) => {
-  await map(context.split('\n').filter(Boolean), context =>
+export const setCommitStatus = async ({ sha, context, state, description, target_url, skip_if_already_set }: SetCommitStatus) => {
+  await map(context.split('\n').filter(Boolean), async context => {
+    if (skip_if_already_set === 'true') {
```
danadajian, line 32: I think we can simplify to this since you're either omitting this or passing `true`
```suggestion
    if (skip_if_already_set) {
```

src/helpers/set-commit-status.ts
```diff
@@ -23,17 +24,33 @@ export class SetCommitStatus extends HelperInputs {
...
+      });
+      const run = check_runs.data.check_runs.find(({ name }) => name === context);
+      if (run) {
+        const runCompletedAndIsValid = run.status === 'completed' && (run.conclusion === 'failure' || run.conclusion === 'success');
+        if (runCompletedAndIsValid) {
+          core.info(`${context} already completed with a ${run.conclusion} conclusion.`);
+          return;
+        }
+      }
+    }
```
danadajian, line 45:
```suggestion
      const runCompletedAndIsValid = run?.status === 'completed' && (run?.conclusion === 'failure' || run?.conclusion === 'success');
      if (runCompletedAndIsValid) {
        core.info(`${context} already completed with a ${run.conclusion} conclusion.`);
        return;
      }
```

src/helpers/set-commit-status.ts
aschwenn: technically yeah, but allowing `'false'` to evaluate to `true` here seems kinda weird

danadajian (approved):

PR #323, opened 2023-01-25
fix(create-pr-comment): create initial comment when none is present

bensbigolbeard:
### :pencil: Description
- Previously, when `login` is provided, it was assumed that there was an existing comment to update. This is not always the case, so it would never work. This change leaves the check for an existing comment from a given user, but now it will create a the initial comment if none exist.
- This also accommodates when both the sha and login are supplied, attempting to update a given comment in a given PR, else creating a new comment

### :link: Related Issues
-


danadajian (approved):

bensbigolbeard:
checking the prettier issues

bensbigolbeard:
i dunno if the failing check is pertinent ...? seems to be some token resolution issue

bensbigolbeard:
ok i was having an issue with the workflow trying to run against my fork, instead of against the repo. looks like all of the workflows are set up with `pull_request` scope, rather than `pull_request_target`. i can put together a separate change to update those. i'll remove the commit with the workflow change from this PR

here was the passing checks with the workflow change
Screen Shot 2023-01-25 at 4 47 35 PM

aschwenn (approved):

bensbigolbeard:
looks like one of you needs to add some PR labels before this is mergeable @danadajian @aschwenn 👍

aschwenn:
> looks like one of you needs to add some PR labels before this is mergeable @danadajian @aschwenn 👍

actually i think it's failing because it's coming from a fork lol

PR #324, opened 2023-01-26
fix: allow changes from forks to pass checks

bensbigolbeard:
### :pencil: Description
- 

### :link: Related Issues
-

### DIFF
--- a/.github/workflows/add-labels.yml
+++ b/.github/workflows/add-labels.yml
@@ -1,8 +1,8 @@
 name: Add Labels
 
 on:
-  pull_request:
-    branches: [ main ]
+  pull_request_target:
+    branches: [main]
     paths:
       - 'src/helpers/add-labels.ts'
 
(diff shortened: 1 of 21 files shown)


.github/workflows/add-labels.yml
```diff
@@ -1,8 +1,8 @@
 name: Add Labels
 
 on:
-  pull_request:
-    branches: [ main ]
+  pull_request_target:
```
danadajian, line 4: So we used to do this. However, `pull_request_target` is a security risk. It exposes all secrets present in workflow, among other [bad things](https://securitylab.github.com/research/github-actions-preventing-pwn-requests/). 🙈 

Thus we had to go back to using `pull_request` in https://github.com/ExpediaGroup/github-helpers/pull/232

danadajian:
So we used to do this. However, `pull_request_target` is a security risk. It exposes all secrets present in workflow, among other [bad things](https://securitylab.github.com/research/github-actions-preventing-pwn-requests/). 🙈 

Thus we had to go back to using `pull_request` in https://github.com/ExpediaGroup/github-helpers/pull/232

bensbigolbeard:
oh dammit. i had this flipped in my brain. i was thinking `pull_request` was the less restrictive one.

PR #348, opened 2023-03-09
fix(delete-stale-branches): prevent default branch deletion

danadajian:
### :pencil: Description
- If a repo somehow has no branch protections, this could actually delete the default branch, which we never want!


aschwenn (approved):
good catch!

PR #359, opened 2023-04-10
feat: is-user-in-team

aschwenn:
### :pencil: Description
- checks if user exists in github organization team

image

image


src/helpers/is-user-in-team.ts
```diff
@@ -0,0 +1,37 @@
...
+
+export class IsUserInTeam extends HelperInputs {
+  login = '';
+  team = '';
+}
+
+export const isUserInTeam = async ({ login = context.actor, team }: IsUserInTeam) => {
+  if (!team) {
+    core.error('Must supply team');
+  }
```
danadajian, line 28: We already handle missing required inputs at the top level

src/helpers/is-user-in-team.ts
```diff
@@ -0,0 +1,37 @@
...
+  const response: OctokitResponse<Record<string, string | number | boolean>[]> = await octokitRequest(
+    'GET /orgs/{org}/teams/{team}/members',
```
danadajian, line 30: Can we not use octokit for this?

aschwenn: annoyingly, octokit doesn't have a wrapper for this (unless you can find one-- i couldn't). the closest thing i see is `octokit.rest.teams.getMembershipForUserInOrg` which 404's if the user isn't in the team or `octokit.rest.teams.getByName` which doesn't include members in the response

aschwenn:
> We could also reuse the existing "teams" input and just return true if the login is a member of at least one of the teams provided

i thought about this but thought it might be unclear as to whether it would be an `and` or `or` operation between the teams. lemme know what you think

src/helpers/is-user-in-team.ts
danadajian: Try teams.listMembersInOrg, which is what we use in a different helper

danadajian (approved):

PR #406, opened 2023-08-09
feat(create-pr): add option to output all data for newly-created PR

aschwenn:
### :pencil: Description
- outputs whole PR object instead of just number

i have a use case where i want the link and title of the newly-created PR

### :link: Related Issues
-


src/helpers/approvals-satisfied.ts
```diff
@@ -50,7 +50,7 @@ export const approvalsSatisfied = async ({ teams, number_of_reviewers = '1', pul
     });
     const codeOwnerLogins = teamsAndLoginsLists.flat().map(({ login }) => login);
 
-    const numberOfCollectiveApprovalsAcrossTeams = approverLogins.filter(login => codeOwnerLogins.includes(login)).length;
+    const numberOfCollectiveApprovalsAcrossTeams = approverLogins.filter(login => login && codeOwnerLogins.includes(login)).length;
```
danadajian, line 53: I think this would only matter if GitHub sent us an empty string login, but this is fine

src/utils/get-core-member-logins.ts
```diff
@@ -29,7 +29,7 @@ export const getCoreMemberLogins = async (pull_number: number, teams?: string[])
 export const getRequiredCodeOwnersEntries = async (pull_number: number): Promise<CodeOwnersEntry[]> => {
   const codeOwners = (await loadOwners(process.cwd())) ?? [];
   const changedFilePaths = await getChangedFilepaths(pull_number);
-  return changedFilePaths.map(filePath => matchFile(filePath, codeOwners)).filter(Boolean);
+  return changedFilePaths.map(filePath => matchFile(filePath, codeOwners)).filter(Boolean) as CodeOwnersEntry[];
```
danadajian, line 32: You removed so many casts, but added this one! 😂 What for?

src/helpers/create-pr.ts
```diff
@@ -26,17 +26,15 @@ export class CreatePR extends HelperInputs {
...
+  const { data } = await octokit.pulls.create({
     title,
     head,
     base: pr_base,
     body,
     maintainer_can_modify: true,
     ...context.repo
   });
-  return number;
+  return data;
```
danadajian, line 37: Could we maybe leverage the `pull_number` input to avoid making this breaking? Like if pull_number is "false" return the whole data object? idk just spitballing

src/helpers/approvals-satisfied.ts
aschwenn: yeah it was just to get the build to pass

src/utils/get-core-member-logins.ts
aschwenn: typescript isn't understanding that the result of this is actually a `CodeOwnersEntry[]` and not a `(CodeOwnersEntry | null)[]`. i hate typecasting but i also didn't want to rewrite this very readable line to placate ts

src/helpers/create-pr.ts
aschwenn: the number will always exist as long as it succeeds, the issue is just that we arbitrarily chose to return the pull number rather than the whole pull object, which would probably make more sense

danadajian: I mean the `pull_number` input to the action, so the consumer can choose to get back the whole data object, and leave the default behavior alone. If this is too jank, we can maybe just make the breaking change but not release a v2 (I'd like to avoid doing that)

aschwenn: ah no this is a good idea. i'll do that. maybe it would be more clear to use a `return_pull_number` input?

danadajian: Maybe something more generic like `return_full_payload`?

aschwenn: sure

danadajian (approved):

PR #443, opened 2023-09-26
feat: create comment-issue-due-date helper

k-borley:
### :pencil: Description
- When an issue is labeled with a priority (critical, high, med, low), this automatically adds a comment with the due date (ex: `This issue is due on Mon Dec 25 2023`). The due date is based on SLA guidelines (see table).

Priority | Days For Resolution
--- | :---:
1 | 2
2 | 14
3 | 45
4 | 90


src/helpers/comment-issue-due-date.ts
```diff
@@ -0,0 +1,61 @@
...
+    case PRIORITY_2:
+      numDaysAllowedOpen = 14;
+      break;
+    case PRIORITY_3:
+      numDaysAllowedOpen = 45;
+      break;
+    case PRIORITY_4:
+      numDaysAllowedOpen = 90;
+      break;
+  }
```
danadajian, line 41: This could probably be a map (i.e. an object mapping)

src/helpers/comment-issue-due-date.ts
```diff
@@ -0,0 +1,61 @@
...
+
+  if (!numDaysAllowedOpen) return;
```
danadajian, line 43: Make `label` a required input rather than returning here

test/helpers/comment-issue-due-date.test.ts
```diff
@@ -0,0 +1,120 @@
...
+        labels: [PRIORITY_4]
+      }
+    });
+    await commentIssueDueDate({ label: PRIORITY_4 });
+    expect(octokit.issues.createComment).toHaveBeenCalledWith({
+      body: 'This issue is due on Mon Dec 25 2023',
+      issue_number: 123,
+      ...context.repo
+    });
+  });
```
danadajian, line 94: These are all one test in my opinion. Just have one test that tests the happy path and one for the label not matching

.github/workflows/comment-issue-due-date.yml
```diff
@@ -0,0 +1,21 @@
...
+on:
+  issues:
+    types:
+      - labeled
+    paths:
+      - 'src/helpers/comment-issue-due-date.ts'
+
+jobs:
+  test:
+    if: contains(fromJSON('["Priority&#58; Critical", "Priority&#58; High", "Priority&#58; Medium", "Priority&#58; Low"]'), github.event.label.name)
```
danadajian, line 12: I think it would be better to put this logic inside the helper so it's more readable

src/helpers/comment-issue-due-date.ts
```diff
@@ -0,0 +1,55 @@
...
+limitations under the License.
+*/
+
+import { HelperInputs } from '../types/generated';
+import { context } from '@actions/github';
+import { octokit } from '../octokit';
+import { PRIORITY_1, PRIORITY_2, PRIORITY_3, PRIORITY_4 } from '../constants';
+
+export class CreateIssueDuedateComment extends HelperInputs {
+  label!: typeof PRIORITY_1 | typeof PRIORITY_2 | typeof PRIORITY_3 | typeof PRIORITY_4;
```
danadajian, line 20:
```suggestion
  label!: keyof typeof daysOpenBasedOnPriority
```

src/helpers/comment-issue-due-date.ts
```diff
@@ -0,0 +1,55 @@
...
+export type PriorityDueDateMapping = {
+  [key: string]: number;
+};
+
+const daysOpenBasedOnPriority: PriorityDueDateMapping = {
+  [PRIORITY_1]: 2,
+  [PRIORITY_2]: 14,
+  [PRIORITY_3]: 45,
+  [PRIORITY_4]: 90
+};
```
danadajian, line 32:
```suggestion
const daysOpenBasedOnPriority = {
  [PRIORITY_1]: 2,
  [PRIORITY_2]: 14,
  [PRIORITY_3]: 45,
  [PRIORITY_4]: 90
} as const;
```

test/helpers/comment-issue-due-date.test.ts
```diff
@@ -0,0 +1,94 @@
...
+  context: { repo: { repo: 'repo', owner: 'owner' }, issue: { number: 123 } },
+  getOctokit: jest.fn(() => ({
+    rest: { issues: { get: jest.fn(), createComment: jest.fn() } }
+  }))
+}));
+
+jest.spyOn(Date, 'now').mockImplementation(() => new Date('2023-09-26T10:00:00Z').getTime());
+
+describe('commentIssueDueDate', () => {
+  it('should add due dates correctly based on priority label', async () => {
```
danadajian, line 31: We should add a test for if an invalid label is added

PR #460, opened 2023-10-16
fix(check-merge-safety): only set merge safety status to failure if it was previous success

aschwenn:
### :pencil: Description
GitHub has a limit of 1000 statuses that can be set per commit. Since `check-merge-safety` runs so many times on large-scale repos, it lately has ended up accidentally hitting this limit, which fails the rest of the job so it doesn't finish setting statuses. Here, I check for an existing failure status before attempting to set a failure status. We could also perform the same check for success statuses, but that's not really a big problem right now; the issue is stale branches that are having the failure status set a thousand times.

### :link: Related Issues
-


src/helpers/check-merge-safety.ts
```diff
@@ -21,6 +21,7 @@ import { map } from 'bluebird';
 import { setCommitStatus } from './set-commit-status';
 import * as core from '@actions/core';
 
+const maxBranchNameLength = 50;
```
danadajian, line 24: Just curious, did the linter get mad at defining this after usage?

src/helpers/check-merge-safety.ts
```diff
@@ -61,6 +75,18 @@ const handlePushWorkflow = async (inputs: CheckMergeSafety) => {
...
+const checkForExistingFailureStatus = async (pullRequest: PullRequest, context: string) => {
+  const { data } = await octokit.repos.getCombinedStatusForRef({
+    ...githubContext.repo,
+    ref: pullRequest.head.sha
+  });
+  if (data.state === 'failure') {
+    const existingContext = data.statuses.find(status => status.context === context);
+    return Boolean(existingContext);
+  }
+  return false;
```
danadajian, line 87: I see the micro-optimization here, but this seems simpler. Am I missing something?
```suggestion
  return Boolean(data.statuses.find(status => status.context === context));
```

src/helpers/check-merge-safety.ts
aschwenn: nah didn't see anything come up

src/helpers/check-merge-safety.ts
aschwenn: statuses will include `Merge Safety` even if the state is `success`, so we need to ensure we're looking at a `failure` before deciding not to re-set the status to `failure`. for a `failure`, failed statuses are listed, and for a `success`, all manually-set successful statuses are listed (afaik)

danadajian: ah ok, makes sense!

src/helpers/check-merge-safety.ts
```diff
@@ -44,13 +45,26 @@ export const checkMergeSafety = async (inputs: CheckMergeSafety) => {
...
   const { state, message } = await getMergeSafetyStateAndMessage(pullRequest, inputs);
-  await setCommitStatus({
-    sha: pullRequest.head.sha,
-    state,
-    context,
-    description: message,
-    ...githubContext.repo
-  });
+  const hasExistingFailureStatus = await checkForExistingFailureStatus(pullRequest, context);
+  if (hasExistingFailureStatus && state === 'failure') {
```
danadajian, line 49: Isn't this `state === 'failure'` check already happening inside `checkForExistingFailureStatus`?

aschwenn: `state` here is the current status, whereas the value from `checkForExistingFailureStatus` is coming from the previous status

basically added this check in to say that in the case where there is an existing `failure` status, only skip if the new status is also a `failure` (it's almost impossible for the status to flip from from `failure` to `success` unless someone were to force-push the target branch, just wanted to make sure our bases were covered)

danadajian: Gotcha - I was getting confused between this state and the individual PR "combined" state

danadajian (approved):

PR #482, opened 2023-11-06
fix(manage-issue-due-dates): remove due soon label if present when adding overdue label

k-borley:
### :pencil: Description
- When adding an "overdue" label, removes the "due soon" label from the issue. No error is thrown if for whatever reason the issue did not have a "due soon label" (was removed manually, issue became overdue before this helper was added to the repo, etc).


src/helpers/manage-issue-due-dates.ts
```diff
@@ -56,6 +56,9 @@ export const manageIssueDueDates = async ({ days = '7' }: ManageIssueDueDates) =
       });
     }
     if (labelToAdd) {
+      if (labelToAdd === OVERDUE_ISSUE) {
+        await octokit.issues.removeLabel({ name: ALMOST_OVERDUE_ISSUE, issue_number, ...context.repo });
```
danadajian, line 60: This will throw an error if the label doesn't exist, right?

danadajian (approved):

PR #546, opened 2024-02-15
feat(manage-merge-queue): do not allow ready for merge when PR is not core approved

PoojaDonekal1294:
### :pencil: Description
Requires PRs to have the `READY_FOR_MERGE` label in order to be added to the merge queue


src/helpers/manage-merge-queue.ts
```diff
@@ -42,6 +43,10 @@ export const manageMergeQueue = async ({ login, slack_webhook_url }: ManageMerge
     core.info('This PR is not in the merge queue.');
     return removePrFromQueue(pullRequest);
   }
+  if (!pullRequest.labels.find(label => label.name === CORE_APPROVED_PR_LABEL)) {
+    core.info('This PR is not core approved.');
```
danadajian, line 47: I wonder if leaving a comment on the PR using [createPrComment](https://github.com/ExpediaGroup/github-helpers/blob/main/src/helpers/create-pr-comment.ts#L52) would be better than just logging in the job

danadajian (commented):
Great idea here! I’m worried that using the `CORE APPROVED` label as a requirement is too assuming, since not all repos necessarily use this. I wonder if we could instead use the [approvalsSatisfied](https://github.com/ExpediaGroup/github-helpers/blob/main/src/helpers/approvals-satisfied.ts#L30) function, which determines if a PR has satisfied core approvals from the required groups.

src/helpers/manage-merge-queue.ts
PoojaDonekal1294: That makes sense! Will do that.

test/helpers/manage-merge-queue.test.ts
```diff
@@ -69,6 +73,33 @@ describe('manageMergeQueue', () => {
...
+          labels: [{ name: READY_FOR_MERGE_PR_LABEL }, { name: 'QUEUED FOR MERGE #1' }]
+        }
+      }));
+      (approvalsSatisfied as jest.Mock).mockResolvedValue(false);
+      await manageMergeQueue();
+    });
+
+    it('should check for approvals satisfied', () => {
+      expect(approvalsSatisfied).toHaveBeenCalledWith();
+    });
```
danadajian, line 96: Instead of this assertion, let's assert that `manageMergeQueue()` is not called

PoojaDonekal1294: But that's the function we are calling, right? Did you mean `setCommitStatus` not being called instead?

danadajian: Oh whoops 🤦 yes sorry!

danadajian (approved):
Nice work!

PR #574, opened 2024-03-26
fix(manage-merge-queue): fix PR removal from queue

danadajian:
### :pencil: Description
- Should fix an issue where the `removePrFromQueue()` function was not being called and thus PRs were not being removed from the queue in certain scenarios.


src/helpers/manage-merge-queue.ts
```diff
@@ -116,7 +116,7 @@ const getQueuedPullRequests = async (): Promise<PullRequestList> => {
 
 export const enableAutoMerge = async (pullRequestId: string, mergeMethod = 'SQUASH') => {
   try {
-    return await octokitGraphql(`
+    await octokitGraphql(`
```
PoojaDonekal1294, line 119: Why are we removing `return` here?

danadajian: Just wanted to remove cases of `return await` since it's kind of confusing to me 😂

PoojaDonekal1294: If we are not using the response returned, then yeah getting rid of `return` makes sense.

PR #581, opened 2024-04-01
feat(approvals-satisfied): add users support

vsingal-p:
### :pencil: Description
- Add users support in codeowners
- Add input users support

### :link: Related Issues
-


src/helpers/approvals-satisfied.ts
```diff
@@ -65,4 +65,14 @@ export const approvalsSatisfied = async ({ teams, number_of_reviewers = '1', pul
   return booleans.every(Boolean);
 };
 
-const createArtificialCodeOwnersEntry = (teams: string[]) => teams.map(team => ({ owners: [team] }));
+const createArtificialCodeOwnersEntry = (teams?: string[], users?: string[]) => [{ owners: (teams || []).concat(users || []) }];
```
danadajian, line 68: Could we destructure the arguments like this?
```suggestion
const createArtificialCodeOwnersEntry = ({ teams = [], users = [] }: { teams?: string[], users?: string[] }) => [{ owners: teams.concat(users) }];
```

src/helpers/approvals-satisfied.ts
```diff
@@ -65,4 +65,14 @@ export const approvalsSatisfied = async ({ teams, number_of_reviewers = '1', pul
...
+const distinct = (arrayWithDuplicates: string[]) => arrayWithDuplicates.filter((n, i) => arrayWithDuplicates.indexOf(n) === i);
+const isTeam = (teamOrUser: string) => teamOrUser.includes('/');
```
danadajian, line 70: I don't think this is always a safe assumption, since you can pass teams without the full organization path. Could we handle teams and users separately?

src/helpers/approvals-satisfied.ts
vsingal-p: I have very less experience in typescript :P
If you say the above works, then it's fine 💯  UTs will verify anyway if it's working on it

src/helpers/approvals-satisfied.ts
vsingal-p: Afaik, teams have to have full path which includes org

Also we cannot handle separately because for manual inputs we can still write separate code, but for codeowners it won't work. So we need a solution which works for both manual input and codeowners file

I believe this will

We can add it in the doc that you should use full path for teams which include org. What say?

danadajian: I'm saying when you build the `loginsLists` list, rather than iterating over teams and users and checking "is team", why not build each list separately and concatenate at the end

vsingal-p: Yes we can do this, but what if the inputs are not provided, and the flow uses codeowners file and within codeowners file users are provided instead of teams?

vsingal-p: As per discussion, this check is fine since CODEOWNERS expect full team name, and in case of teams in input we are updating and validating it

src/helpers/approvals-satisfied.ts
```diff
@@ -65,4 +73,40 @@ export const approvalsSatisfied = async ({ teams, number_of_reviewers = '1', pul
...
+    return teamsList.map(team => {
+      if (!team.includes('/')) {
+        return `${context.repo.owner}/${team}`;
+      } else {
+        return team;
+      }
+    });
+  } else {
+    return undefined;
+  }
```
danadajian, line 100:
```suggestion
    return teamsList?.map(team => {
      if (!team.includes('/')) {
        return `${context.repo.owner}/${team}`;
      } else {
        return team;
      }
    });
```

src/helpers/approvals-satisfied.ts
```diff
@@ -65,4 +73,40 @@ export const approvalsSatisfied = async ({ teams, number_of_reviewers = '1', pul
...
+
+const validateTeamsList = (teamsList?: string[]) => {
+  if (teamsList) {
+    return teamsList.every(team => {
+      const inputOrg = team.split('/')[0];
+      return inputOrg === context.repo.owner;
+    });
+  } else {
+    return true;
+  }
```
danadajian, line 111:
```suggestion
  return teamsList?.every(team => {
      const inputOrg = team.split('/')[0];
      return inputOrg === context.repo.owner;
    }) ?? true;
```

test/helpers/approvals-satisfied.test.ts
```diff
@@ -3,7 +3,7 @@ Copyright 2021 Expedia, Inc.
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
 You may obtain a copy of the License at
-    https://www.apache.org/licenses/LICENSE-2.0
+    https://www.apache.licenses/LICENSE-2.0
```
danadajian, line 6:
```suggestion
    https://www.apache.org/licenses/LICENSE-2.0
```

danadajian (approved):

PR #584, opened 2024-04-04
feat: update close-pr to support other repos, etc

amandaolsen:
Made these changes with @aschwenn.

### :pencil: Description
- Add more abilities to the close-pr action.

### :link: Related Issues
-


aschwenn (approved):
hopped on a call with @amandaolsen and we tested this new functionality and verified that it actually works between real repos!

PR #591, opened 2024-04-10
feat(generate-matrix): add optional load balancing

aschwenn:
### :pencil: Description
adding optional load balancing into `generate-matrix`! this is intended to prevent individual matrix jobs from having a way higher computational load than others. the tradeoff with the algorithm was either a fully optimized load balance but losing input paths order, or retaining input paths order but not 100% optimizing load balancing (and having possibly small variance in the outputted batch number as compared to the `batches` input integer). i chose the latter, since otherwise trying to figure out what batch a certain path was placed in in the github actions logs would be a nightmare, and i don't think the load balancing needs to be _perfect_ to be an improvement.

the test cases will expand on the intended usage, but here's a quick snippet:

```
- name: Generate path matrix
  id: paths
  uses: ExpediaGroup/github-helpers@v1
  with:
    helper: generate-matrix
    paths: package/a,package/b,package/c,package/d
    load_balancing_sizes: 3,1,1,1
    batches: 2
```

which results in

```
{
  "includes": [
    { "path": "package/a" },
    { "path": "package/b,package/c,package/d" }
  ]
}
```

### :link: Related Issues
-


src/helpers/generate-matrix.ts
```diff
@@ -17,11 +19,42 @@ import { chunk } from 'lodash';
...
+  } else {
+    const loadBalancingSizes = load_balancing_sizes.split(/[\n,]/).map(size => Number(size));
+    if (loadBalancingSizes.length !== matrixValues.length) throw new Error('load_balancing_sizes have the same length as paths');
+    const targetLoadSize = loadBalancingSizes.reduce((acc, size) => acc + size, 0) / batches;
+    const loadBalancedPaths: string[] = [];
+    let currentLoadSize = 0;
+    let currentBatch: string[] = [];
+    matrixValues.forEach((path, index) => {
+      const possibleLoadSize = currentLoadSize + loadBalancingSizes[index];
+      if (Math.abs(possibleLoadSize - targetLoadSize) <= Math.abs(loadBalancingSizes[index] - targetLoadSize)) {
```
aschwenn, line 41: this is meant to push following loads into the next batch in the case that they result in the current load far exceeding the `targetLoadSize` (for example, without this, with a `targetLoadSize` of `5` and a `currentLoadSize` of `4`, a new load size of `20` would be lumped into the current batch. we obviously would like to avoid that, since such an outlier should be in its own batch)

i can put a comment in the code if that would be helpful for maintainability

src/helpers/generate-matrix.ts
```diff
@@ -17,11 +19,42 @@ import { chunk } from 'lodash';
...
   const matrixValues = paths.split(/[\n,]/);
-  return {
-    include: chunk(matrixValues, Math.ceil(matrixValues.length / Number(batches))).map(chunk => ({ path: chunk.join(',') }))
-  };
+  const batches = Number(_batches);
+  if (!load_balancing_sizes) {
+    return {
+      include: chunk(matrixValues, Math.ceil(matrixValues.length / batches)).map(chunk => ({ path: chunk.join(',') }))
+    };
+  } else {
```
danadajian, line 32: Unnecessary else block here, can just use the [guard clause pattern](https://deviq.com/design-patterns/guard-clause)

src/helpers/generate-matrix.ts
```diff
@@ -17,11 +19,42 @@ import { chunk } from 'lodash';
...
+    const loadBalancingSizes = load_balancing_sizes.split(/[\n,]/).map(size => Number(size));
+    if (loadBalancingSizes.length !== matrixValues.length) throw new Error('load_balancing_sizes have the same length as paths');
```
danadajian, line 34:
```suggestion
    if (loadBalancingSizes.length !== matrixValues.length) throw new Error('load_balancing_sizes input must have the same length as paths input');
```

src/helpers/generate-matrix.ts
```diff
@@ -17,11 +19,42 @@ import { chunk } from 'lodash';
...
+    const targetLoadSize = loadBalancingSizes.reduce((acc, size) => acc + size, 0) / batches;
+    const loadBalancedPaths: string[] = [];
+    let currentLoadSize = 0;
+    let currentBatch: string[] = [];
+    matrixValues.forEach((path, index) => {
```
danadajian, line 39: Wonder if we can map the `matrixValues` to `loadBalancedPaths` rather than iterating and mutating

test/helpers/generate-matrix.test.ts
```diff
@@ -36,4 +36,70 @@ describe('generateMatrix', () => {
       ]
     });
   });
+
+  it('should generate matrix json with appropriate load balancing in naive case', () => {
```
danadajian, line 40: Might be good to explain what naive, moderate, and advanced cases are

danadajian:
> the tradeoff with the algorithm was either a fully optimized load balance but losing input paths order, or retaining input paths order but not 100% optimizing load balancing (and having possibly small variance in the outputted batch number as compared to the batches input integer)

Also curious what you mean by this

test/helpers/generate-matrix.test.ts
aschwenn: a better demarcation would be `small | medium | large`

src/helpers/generate-matrix.ts
aschwenn: a more functional approach would be to use `reduce` instead of `forEach`, but my worry was that this would be less readable/maintainable. i don't see a great way to compute the result in-place but it could be done with an in-place `number[][]` of loads and a separate `string[]` of resulting paths initialized before the `reduce` block. then there wouldn't be any reassignment

aschwenn:
> > the tradeoff with the algorithm was either a fully optimized load balance but losing input paths order, or retaining input paths order but not 100% optimizing load balancing (and having possibly small variance in the outputted batch number as compared to the batches input integer)
> 
> Also curious what you mean by this

for example, if we had an input of

```
- name: Generate path matrix
  id: paths
  uses: ExpediaGroup/github-helpers@v1
  with:
    helper: generate-matrix
    paths: package/a,package/b,package/c,package/d
    load_balancing_sizes: 2,4,3,1
    batches: 2
```

a 100% load balanced result would be `package/a,package/c`, and `package/b,package/d`, since both batches would have a cumulative load of `5` which it the total load divided by the desired batch size. but now we've played around with the order and it can become more difficult to find which job may contain a certain path, since they're no longer in the supplied order. i think retaining order is probably more valuable than having the more accurate load balancing

this can also result in situations sometimes where we don't have the exact number of `batches` in the output as we specified in the input, since we're no longer dividing the quantity of paths evenly by batches. so if i specify `batches: 4`, it could have an edge case where the helper actually returns 5 batches. think this is a dealbreaker?

danadajian:
> > > the tradeoff with the algorithm was either a fully optimized load balance but losing input paths order, or retaining input paths order but not 100% optimizing load balancing (and having possibly small variance in the outputted batch number as compared to the batches input integer)
> 
> > 
> 
> > Also curious what you mean by this
> 
> 
> 
> for example, if we had an input of
> 
> 
> 
> ```
> 
> - name: Generate path matrix
> 
>   id: paths
> 
>   uses: ExpediaGroup/github-helpers@v1
> 
>   with:
> 
>     helper: generate-matrix
> 
>     paths: package/a,package/b,package/c,package/d
> 
>     load_balancing_sizes: 2,4,3,1
> 
>     batches: 2
> 
> ```
> 
> 
> 
> a 100% load balanced result would be `package/a,package/c`, and `package/b,package/d`, since both batches would have a cumulative load of `5`, which is the total load divided by the desired batch size. but now we've played around with the order and it can become more difficult to find which job may contain a certain path, since they're no longer in the supplied order. i think retaining order is probably more valuable than having the more accurate load balancing
> 
> 
> 
> this can also result in situations sometimes where we don't have the exact number of `batches` in the output as we specified in the input, since we're no longer dividing the quantity of paths evenly by batches. so if i specify `batches: 4`, it could have an edge case where the helper actually returns 5 batches. think this is a dealbreaker?

Not a dealbreaker. I think this makes sense as a first iteration!

This also makes me wonder if there is some library out there that we could outsource this logic to. Like given a list of numbers, pick n groups with as close to a given batch size as possible, with order not mattering

src/helpers/generate-matrix.ts
```diff
@@ -11,17 +11,51 @@ See the License for the specific language governing permissions and
...
+  const batches = Number(_batches);
+  if (!load_balancing_sizes) {
+    return {
+      include: chunk(matrixValues, Math.ceil(matrixValues.length / batches)).map(chunk => ({ path: chunk.join(',') }))
+    };
+  }
+  const loadBalancingSizes = load_balancing_sizes.split(/[\n,]/).map(size => Number(size));
+  if (loadBalancingSizes.length !== matrixValues.length)
+    throw new Error('load_balancing_sizes input must have the same length as paths input');
+  const targetLoadSize = loadBalancingSizes.reduce((acc, size) => acc + size, 0) / batches;
```
danadajian, line 37: Nit pick - we could probably use the `sum` function from lodash to make this slightly more readable

danadajian (approved):

src/helpers/generate-matrix.ts
aschwenn: sure that makes sense

aschwenn:
> Not a dealbreaker. I think this makes sense as a first iteration!
> 
> This also makes me wonder if there is some library out there that we could outsource this logic to. Like given a list of numbers, pick n groups with as close to a given batch size as possible, with order not mattering

yeah i tried looking, most things with `load balancing` keywords were infra or server-related. i'd imagine there's a leetcode problem out there that optimizes the exact problem i'm trying to solve here. i'll keep an eye out as we continue to test things

PR #631, opened 2024-07-09
feat(approvals-satisfied): comment when PR does not meet approvals

sjschmidt93:
### :pencil: Description
- 

### :link: Related Issues
-


src/helpers/approvals-satisfied.ts
```diff
@@ -80,15 +83,27 @@ export const approvalsSatisfied = async ({
...
+
   const booleans = await Promise.all(requiredCodeOwnersEntriesWithOwners.map(codeOwnersEntrySatisfiesApprovals));
-  return booleans.every(Boolean);
+  const approvalsSatisfied = booleans.every(Boolean);
+
+  core.info(logsJoined);
+
+  if (!approvalsSatisfied) {
+    await createPrComment({
+      body: 'PRs must meet all required approvals before entering the merge queue.\n\n' + logsJoined
```
danadajian, line 102: This helper doesn't only run in the context of the merge queue, so we should make the message more generic. Maybe like 
```
Required approvals not satisfied: <logsJoined>
```
and we could pass the existing `body` input into this helper so the merge queue invocation can provide a merge queue-specific message too

src/helpers/approvals-satisfied.ts
```diff
@@ -80,15 +80,32 @@ export const approvalsSatisfied = async ({
...
+  logs.push(`Required code owners: ${requiredCodeOwnersEntriesWithOwners.map(({ owners }) => owners).toString()}`);
+
   const booleans = await Promise.all(requiredCodeOwnersEntriesWithOwners.map(codeOwnersEntrySatisfiesApprovals));
-  return booleans.every(Boolean);
+  const approvalsSatisfied = booleans.every(Boolean);
+
+  core.info(logs.join('\n'));
+
+  if (!approvalsSatisfied) {
+    logs.unshift('Required approvals not satisfied:\n');
```
sjschmidt93, line 97: never heard of unshift until now, but it is equal to `['Required approvals not satisifed', ...logs]`

src/helpers/approvals-satisfied.ts
```diff
@@ -30,13 +31,10 @@ export class ApprovalsSatisfied extends HelperInputs {
...
-export const approvalsSatisfied = async ({
-  teams,
-  users,
-  number_of_reviewers = '1',
-  required_review_overrides,
-  pull_number
-}: ApprovalsSatisfied = {}) => {
+export const approvalsSatisfied = async (
+  { teams, users, number_of_reviewers = '1', required_review_overrides, pull_number }: ApprovalsSatisfied = {},
+  approvalsNotMetMessage: string | undefined = undefined
```
danadajian, line 36: Can we use the existing `body` input instead of this second argument for providing an optional comment body? `approvalsSatisfied` is a standalone action

src/helpers/approvals-satisfied.ts
```diff
@@ -21,21 +21,24 @@ import { CodeOwnersEntry } from 'codeowners-utils';
...
 }
 
 export const approvalsSatisfied = async ({
   teams,
   users,
   number_of_reviewers = '1',
   required_review_overrides,
-  pull_number
+  pull_number,
+  approvals_not_met_message
```
danadajian: Can we rename to `body` and add this to the above interface?

danadajian (approved):

PR #651, opened 2024-08-07
feat(manage-merge-queue): allow only maintainers to jump the queue

bherrmann2:
### :pencil: Description
- Adding an option to enable only allowing maintainers to be able to jump the queue

### :link: Related Issues
-


src/helpers/is-user-in-team.ts
```diff
@@ -16,7 +16,7 @@ import { context } from '@actions/github';
 import { octokit } from '../octokit';
 
 export class IsUserInTeam extends HelperInputs {
-  login = '';
+  login? = '';
```
bherrmann2, line 19: had to switch this so the default value on line 23 can actually be leveraged

danadajian (approved):

PR #652, opened 2024-08-09
fix(manage-merge-queue): ensure queued for merge label is applied

danadajian:
### :pencil: Description
- Fixes an issue with the merge queue manager where PRs that have `READY FOR MERGE` without a "queued for merge" label screws up queue assignment.


src/helpers/manage-merge-queue.ts
```diff
@@ -63,7 +63,7 @@ export const manageMergeQueue = async ({
     return removePrFromQueue(pullRequest);
   }
   const queuedPrs = await getQueuedPullRequests();
-  const queuePosition = queuedPrs.length;
+  const queuePosition = queuedPrs.length + 1;
```
bherrmann2, line 66: Length isn't an offset. Why the `+ 1`?

danadajian: `getQueuedPullRequests()` will now return a list of open PRs with both the `READY FOR MERGE` and `QUEUED FOR MERGE #X` labels. Before, this was _including_ the PR we are running this for

danadajian: So now, if there are 2 PRs in the queue, and this runs for a new PR that just got `READY FOR MERGE`, `queuedPrs.length` should be 2, and thus `queuePosition` for this new PR should be 3.

PR #674, opened 2024-12-03
feat(get-changed-files): include original location for renamed files in response

aschwenn:
### :pencil: Description
when files are renamed with no changes, github doesn't treat the change as an `addition` + a `deletion`, it treats it as one singular `renamed` object. this isn't the most helpful in a monorepo setting, where we may have a file moved (renamed) from one package to another, in which case we'd want to test both packages, not just the one the file was moved into.

this change pulls in all `previous_filename` fields on files that have been `renamed` to append to the full list. that object coming back from the GH API/octokit looks like:

```json
{
    "sha": "dcc78ba",
    "filename": "new/file/location/file.js",
    "status": "renamed",
    "additions": 4,
    "deletions": 6,
    "changes": 10,
    "blob_url": "https://github.com/...",
    "raw_url": "https://github.com/...",
    "contents_url": "https://api.github.com/repos/...",
    "patch": "...",
    "previous_filename": "previous/file/location/file.js"
  },
```

**to test**: i've added a case to the test file to account for this 🙂


test/helpers/get-changed-files.test.ts
```diff
@@ -118,4 +145,13 @@ describe('getChangedFiles', () => {
...
     expect(result).toEqual(mock_data3[0]!.filename);
   });
+
+  it('should include original location of renamed files, as if rename was an addition and deletion', async () => {
+    (octokit.pulls.listFiles as unknown as Mocktokit).mockImplementation(async ({ page }) => ({
+      data: page === 1 ? mock_data4 : []
+    }));
+    const result = await getChangedFiles({ ignore_deleted: 'true' });
+
+    expect(result).toEqual([mock_data4[0]!.filename, mock_data4[1]!.filename, mock_data4[1]!.previous_filename].join(','));
```
danadajian, line 155: If you put `as const` after `mock_data4` you shouldn't need the non-null assertions

src/utils/get-changed-filepaths.ts
```diff
@@ -17,8 +17,17 @@ import { octokit } from '../octokit';
...
+    changedFiles.reduce((acc, file) => {
+      if (ignore_deleted && file.status === 'removed') return acc;
+      acc.add(file.filename);
+      if (file.status === 'renamed' && file.previous_filename) {
+        acc.add(file.previous_filename);
+      }
+      return acc;
+    }, new Set<string>())
+  );
+  return files;
```
danadajian, line 30: This is pretty hard to read -- could we just create two arrays and concat them?

aschwenn: sure, i was using a hashset to dedupe but in hindsight i don't think there's a possible scenario where there could be duplicates

danadajian (approved):

PR #699, opened 2025-02-06
feat(check-merge-safety): fallback to git command when diff fails

bensbigolbeard:
### :pencil: Description
- adds fallback for diff attempts through the compare endpoint, which fails with "diff too large" errors for large repos with many deltas between SHAs. 
- the change is to first attempt the existing diff call, then when you receive a 406 (diff too large error code), fetch each sha and use the local git installation to run the diff. 
- new failure states: 
  - when diff fails with 40x errors, via `octokit.repos.compareCommitsWithBasehead`, the catch block processes matching cases, else it's thrown again to match current handling
  - when individual branches/SHAs cannot be retrieved via `git.fetch`
  - when diffs between two SHAs fail via `git.diff`

### :link: Related Issues
-


bensbigolbeard:
https://github.com/ExpediaGroup/github-helpers/pull/702

PR #705, opened 2025-02-28
fix(approvals-satisfied): fix teams with over 100 members

treyp:
### :pencil: Description
- In the `approvals-satisfied` script, when fetching the members for a GitHub team, only the first page of 100 members is fetched. This is a limitation for teams with more than 100 members. This PR adds support for fetching all pages of members of an org.

### :link: Related Issues
-


src/utils/paginate-members-in-org.ts
```diff
@@ -0,0 +1,33 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+import { MembersInOrg } from '../types/github';
+import { octokit } from '../octokit';
+import { convertToTeamSlug } from './convert-to-team-slug';
+import { context } from '@actions/github';
+
+export const paginateMembersInOrg = async (team: string, page: number = 1): Promise<MembersInOrg> => {
```
sayali10, line 19: Something to look out for: If someone tried to fetch members of large github teams it might end up hitting API limit.

sayali10 (approved):

src/utils/paginate-members-in-org.ts
danadajian: Looks like the [max team size is 5000](https://docs.github.com/en/enterprise-cloud@latest/organizations/managing-saml-single-sign-on-for-your-organization/managing-team-synchronization-for-your-organization#usage-limits), in which case this would use 500 API hits which wouldn't reach any rate limit.

danadajian (approved):

PR #706, opened 2025-03-04
feat(approvals-satisfied): overrides for CODEOWNERS

treyp:
### :pencil: Description

In the `approvals-satisfied` script, to specify which approvals are needed, the current parameters support either specifying teams + users, or letting CODEOWNERS be the source of truth.

This generally works, but there are times when we'd like a little of both. After all, if CODEOWNERS was flexible enough, we wouldn't need this `approvals-satisfied` GitHub Action in the first place. 

One such case is in large repos. For larger teams, assigning code ownership of a directory to a large team using CODEOWNERS is problematic because every event notifies every member of the team which is automatically requested for review. There is [no way for GitHub users to customize these notifications](https://github.com/orgs/community/discussions/35673). One workaround is to use GitHub Actions like this to handle fine-grained approvals. 

Here, we add support for still reading from CODEOWNERS, but allowing specific overrides of it. For the use case above, this allows us to specify no owner for the directory in CODEOWNERS (defaulting to any user with write permissions counting towards the minimum approvals needed, preventing a review request and thus notifications), but then enforce a specific number of approvals by specific GitHub teams/users using this GitHub Action.


danadajian:
It feels like [team code review settings](https://docs.github.com/en/organizations/organizing-members-into-teams/managing-code-review-settings-for-your-team#disabling-auto-assignment) could solve your notification problem, no?

treyp:
> It feels like [team code review settings](https://docs.github.com/en/organizations/organizing-members-into-teams/managing-code-review-settings-for-your-team#disabling-auto-assignment) could solve your notification problem, no?

Afraid not. We don't use auto-assignment, which picks specific individuals from the team and sets the PR reviewers to them as individuals instead. Instead, the team is set as the reviewer, similar to this PR:

Screenshot 2025-03-03 at 10 10 25 PM

All individuals from the team receive notifications for all such PRs. There is no way to disable them, hence the GitHub feature requests about it.

src/utils/get-core-member-logins.ts
```diff
@@ -26,8 +26,23 @@ export const getCoreMemberLogins = async (pull_number: number, teams?: string[])
...
+      if (!pattern) {
+        core.setFailed(
+          'Invalid code_owners_override format. Please provide a comma-separated list of lines in GitHub CODEOWNERS format. For example, "/foo @owner1 @owner2,/bar @owner3".'
+        );
+        throw new Error();
+      }
+      return { pattern, owners };
+    });
+    // codeowners-utils sorts its CodeOwnersEntry array in reverse order of the CODEOWNERS file
+    codeOwnerOverrides.reverse();
```
danadajian, line 43:
```suggestion
    }).toReversed(); // codeowners-utils sorts its CodeOwnersEntry array in reverse order of the CODEOWNERS file
```

treyp: @danadajian originally, i was going to use `.toReversed()`, but i was worried about Node compatibility -- looks like it [didn't land until Node 20](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/toReversed#browser_compatibility) which may exclude some consumers. happy to chain the `.reverse()` call if you'd prefer though.

danadajian: This action [runs on node 20](https://github.com/ExpediaGroup/github-helpers/blob/main/action.yml#L164), so we should be good

treyp: FWIW, Jest still seems to use Node to run (despite being launched with bun), so if the user's default Node environment is <20, the tests will fail. However, moving forward since this repo's GitHub workflows seem to work fine.

danadajian (approved):

danadajian (changes requested):
Since you're adding a new input you have to add it to action.yml

src/utils/get-core-member-logins.ts
danadajian: Node 18's end of life is [next month](https://github.com/nodejs/Release?tab=readme-ov-file#release-schedule) so honestly I'm not too worried about people using Node 18 😄

danadajian (approved):

PR #717, opened 2025-05-12
feat(delete-deployment): Enhancement to delete prev deployments

jacklu97:
### :pencil: Description
PR enhances the current development of the delete-deployment action making it possible to remove all deployments related to an environment if no sha is given. This increases the flexibility of the action making it useful for extended use cases.

## Changes

- sha is no longer a mandatory parameter, if given will target original behavior
- Action can remove non active deployments linked to a given environment
- Return type specified. Now action will return an object with information about the flow that can be used in next steps of a given workflow

### Response example

Given the class
```typescript
class DeleteDeploymentResponse {
  deploymentsDeleted = 0;
  deploymentsFound = 0;
  message = '';
  environmentDeleted = false;

  constructor(init?: Partial<DeleteDeploymentResponse>) {
    Object.assign(this, init);
  }
}
```

A response can look like
```json5
{
  deploymentsDeleted = 10;
  deploymentsFound = 10;
  message = 'Deleted 10 deployments for environment test';
  environmentDeleted = true;
}
```

## Motivation

Some workflows can have a flow where they deploy different versions of changes on an specific env having a relationship of n deployments per env. Current impl only takes care of current deployment which can lead to have orphan deployments that don't get deleted if the helper was run before.

### :link: Related Issues
-


src/helpers/delete-deployment.ts
```diff
@@ -16,37 +16,78 @@ import { HelperInputs } from '../types/generated';
...
+  const deleteRequests = deployments.map(deleteDeploymentPromise);
+
+  const reqResults = await Promise.allSettled([
+    ...deleteRequests,
+    octokit.repos.deleteAnEnvironment({
+      environment_name: environment,
+      ...context.repo,
+      ...GITHUB_OPTIONS
+    })
+  ]);
```
danadajian, line 79: I prefer using the `Bluebird.map` function which handles multiple promise resolution for us. [Example usage](https://github.com/ExpediaGroup/github-helpers/blob/35d29d5a88cfd640ec4f529bd92432797bdbe434/src/helpers/set-commit-status.ts#L31)

src/helpers/delete-deployment.ts
```diff
@@ -16,37 +16,78 @@ import { HelperInputs } from '../types/generated';
...
+  deploymentsDeleted = 0;
+  deploymentsFound = 0;
+  message = '';
+  environmentDeleted = false;
+
+  constructor(init?: Partial<DeleteDeploymentResponse>) {
+    Object.assign(this, init);
+  }
+}
+
```
danadajian, line 29: Is there a reason to return this arbitrary object? I'd rather log using `core.info()` so users can see the behavior in the action logs

jacklu97: The idea was to return the information in small chunks so users could take it in next steps as needed. Is it possible to use core.output per member of the class so it is not required to stringify the object and users can actually access them by `steps.clean-step.outputs.deploymentsDeleted`?

src/helpers/delete-deployment.ts
jacklu97: Will take a look

src/helpers/delete-deployment.ts
danadajian: Looks like we can only set output to strings
image

jacklu97: We could do something like:
```typescript
for (const [key, value] of Object.entries(myResponseInstance)) {
  core.setOutput(key, value);
}
```

this would be the flexibility of accessing the objects directly. Otherwise the whole object would become a json string and would require an step (whenever this action is used) to read it and extract the desired object.

danadajian: Curious what your use case is to have all this metadata available in the workflow?

jacklu97: We have a workflow that deploys our branch of a PR in a temp env. The thing is after this was deployed, every new push will create a new deployment to the same env. Now, this data in the output is planned to be used in another action to add a comment to the PR informing what was removed

danadajian: It might be easier to support adding a comment in this action using the `sha` input to find the PR. We could do this in a follow-up

src/helpers/delete-deployment.ts
```diff
@@ -15,38 +15,89 @@ import { GITHUB_OPTIONS } from '../constants';
...
-        ...context.repo,
-        ...GITHUB_OPTIONS
-      })
-    ]);
   }
+
+  const deployments = data.map(deployment => deployment.id);
+
+  const deactivateRequests = deployments.map(setDeploymentStatus);
+  await Promise.all(deactivateRequests);
```
danadajian, line 61: Can we use `Bluebird.map` here as well?

src/helpers/delete-deployment.ts
```diff
@@ -15,38 +15,89 @@ import { GITHUB_OPTIONS } from '../constants';
...
+      try {
+        const reqResult = await octokit.repos.deleteDeployment({
+          deployment_id: deploymentId,
+          ...context.repo,
+          ...GITHUB_OPTIONS
+        });
+        return {
+          status: 'fulfilled',
+          value: reqResult
+        };
```
danadajian, line 75: I don't see the value of adding an arbitrary wrapper for the response. Can't we just use the octokit response? It has a `status` field that we can use instead of the `fulfilled` stuff

src/helpers/delete-deployment.ts
jacklu97: If sha is given then it would not delete all orphan deployments of the PR, however I think we can use github context to get the PR the action is linked to and add that comment

src/helpers/delete-deployment.ts
jacklu97: Sure, updated

src/helpers/delete-deployment.ts
jacklu97: Addressed, added a check with the status that comes from actual response

src/helpers/delete-deployment.ts
```diff
@@ -11,42 +11,99 @@ See the License for the specific language governing permissions and
...
+  }
+}
 
 export class DeleteDeployment extends HelperInputs {
-  sha = '';
   environment = '';
 }
 
-export const deleteDeployment = async ({ sha, environment }: DeleteDeployment) => {
+const deactiveDeployments = async (deployments: number[]) => {
```
danadajian, line 38:
```suggestion
const deactivateDeployments = async (deployments: number[]) => {
```

danadajian (approved):

PR #738, opened 2025-08-22
feat(is-user-core-member): overrides for CODEOWNERS

treyp:
### :pencil: Description

Bring the `codeowners_overrides` parameter introduced in #705 and #706 to another helper, this time to the `is-user-core-member` helper.


src/helpers/is-user-core-member.ts
```diff
@@ -18,9 +18,9 @@ import { getCoreMemberLogins } from '../utils/get-core-member-logins';
 
 export class IsUserCoreMember extends HelperInputs {}
 
-export const isUserCoreMember = async ({ pull_number, login = context.actor }: IsUserCoreMember) => {
+export const isUserCoreMember = async ({ pull_number, login = context.actor, codeowners_overrides }: IsUserCoreMember) => {
   const pullNumber = Number(pull_number);
-  const coreMembers = await getCoreMemberLogins(pullNumber);
+  const coreMembers = await getCoreMemberLogins(pullNumber, undefined, codeowners_overrides);
```
danadajian, line 23: Can we refactor this method signature to accept a params object so we don't need to pass undefined?

treyp: @danadajian done! ✅

PR #781, opened 2026-04-23
fix(check-merge-safety): Keep failure description short on comment matches, add target_url to failure statuses

avwu99:
### :pencil: Description

1. Added getWorkflowRunUrl() helper - builds the URL to the current workflow run using githubContext.serverUrl, repo, and runId
3. Added target_url to failure states - for comment paths, global files, and project failures, the commit status now includes a link to the workflow run where full details are logged
5. Simplified comment paths failure message - changed from listing paths (which could exceed 140 char limit) to Branch is behind on N path(s) from comment. Please update with main.

Now when a merge safety check fails, users can click the "Details" link next to the status check to go directly to the workflow logs where the full list of outdated paths is printed via core.error().

### :link: Related Issues
-


src/helpers/check-merge-safety.ts
```diff
@@ -68,6 +71,7 @@ const setMergeSafetyStatus = async (pullRequest: PullRequest, { context = 'Merge
       state,
       context,
       description: message,
+      target_url,
```
danadajian:
```suggestion
      target_url: getWorkflowRunUrl(),
```
Can't we just inline it here?

danadajian (approved):

